PR #1, opened 2020-01-31
Adding @checkup/cli package with scaffold of checkup command.

scalvert:


.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  test:
+    name: Tests
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macOS-latest]
+
```
rwjblue: Probably good to add multiple node versions (and maybe windows too)

packages/cli/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
rwjblue: We should move this into the project root I think (most editor plugins expect it to be a peer of `.git` dir IIRC)

scalvert: Yep. There's a bunch of things that get generated by oclif that are assuming a root level project.

.github/workflows/ci-build.yml
scalvert: Will add nodes 10 and 12

packages/cli/.editorconfig
scalvert: Fixed.

.eslintrc
```diff
@@ -0,0 +1,22 @@
...
+    "ecmaVersion": 2017,
+    "sourceType": "module"
+  },
+  "plugins": ["prettier", "@typescript-eslint"],
+  "extends": ["eslint:recommended", "prettier", "prettier/@typescript-eslint"],
+  "env": {
+    "browser": false,
+    "node": true,
+    "es6": true,
+    "mocha": true
```
rwjblue: Probably should limit `mocha` to the test subdirs

.eslintignore
```diff
@@ -0,0 +1,2 @@
+packages/*/lib
```
rwjblue: Just to confirm, this the local of the compiled output?

package.json
```diff
@@ -0,0 +1,37 @@
...
+  "repository": "https://github.com/checkupjs/checkup.git",
+  "author": "Steve Calvert <steve.calvert@gmail.com>",
+  "license": "MIT",
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ],
+  "scripts": {
+    "test": "yarn workspaces run test",
+    "lint": "eslint packages/* --ext .ts"
```
rwjblue: I think you should probably do:

```suggestion
    "lint": "eslint . --cache --ext .ts"
```

Specifically, we want to cache (makes repeated runs much faster) and we should be able to do `eslint .`, that way we also ensure any top level scripts get linted (like `.eslintrc.js`). 

You may need to tweak `.eslintignore` do enable this.

package.json
```diff
@@ -0,0 +1,37 @@
...
+    "prettier": "^1.19.1"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "packages/**/*.ts": [
+      "eslint packages/* --ext .ts"
```
rwjblue: If you cache this, it will make commiting faster.

Also, why isn't this just calling `yarn lint` (seems annoying to have to include the same snippet in two places)?

packages/cli/bin/run
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env node
+
+const fs = require('fs')
+const path = require('path')
+const project = path.join(__dirname, '../tsconfig.json')
+const dev = fs.existsSync(project)
+
+if (dev) {
```
rwjblue: This relies on the fact that `tsconfig.json` isn't published (which is fine), but can you double check that it is included in `.npmignore`?

rwjblue: Just checked, looks like we are using `files/` in `package.json` instead, so we don't need to do anything special here.

packages/cli/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "author": "Steve Calvert <steve.calvert@gmail.com>",
+  "bin": {
+    "checkup": "./bin/run"
+  },
+  "bugs": "https://github.com/scalvert/https://github.com/checkupjs/checkup/issues",
+  "dependencies": {
+    "@oclif/command": "^1",
+    "@oclif/config": "^1",
+    "@oclif/plugin-help": "^2",
+    "tslib": "^1"
```
rwjblue: What is this used for?

packages/cli/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "typescript": "^3.3"
+  },
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "files": [
+    "/bin",
+    "/lib"
+  ],
+  "homepage": "https://github.com/scalvert/https://github.com/checkupjs/checkup",
```
rwjblue: Seems wrong)

packages/cli/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@checkup/cli",
+  "description": "A CLI that provides health check information about your project.",
+  "version": "0.0.0",
+  "author": "Steve Calvert <steve.calvert@gmail.com>",
+  "bin": {
+    "checkup": "./bin/run"
+  },
+  "bugs": "https://github.com/scalvert/https://github.com/checkupjs/checkup/issues",
```
rwjblue: Seems wrong

packages/cli/README.md
```diff
@@ -0,0 +1,42 @@
+# @checkup/cli
+
+A CLI that provides health check information about your project.
+
+![CI Build](https://github.com/checkupjs/checkup/workflows/CI%20Build/badge.svg)
+[![oclif](https://img.shields.io/badge/cli-oclif-brightgreen.svg)](https://oclif.io)
+[![Version](https://img.shields.io/npm/v/@checkup/cli.svg)](https://npmjs.org/package/@checkup/cli)
+[![Downloads/week](https://img.shields.io/npm/dw/@checkup/cli.svg)](https://npmjs.org/package/@checkup/cli)
+[![License](https://img.shields.io/npm/l/@checkup/cli.svg)](https://github.com/scalvert/https://github.com/checkupjs/checkup/blob/master/package.json)
```
rwjblue: The `github.com/scalvert` path seems wrong

packages/cli/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@types/mocha": "^5",
+    "@types/node": "^10",
+    "chai": "^4",
+    "mocha": "^5",
+    "nyc": "^14",
+    "ts-node": "^8",
+    "typescript": "^3.3"
+  },
+  "engines": {
+    "node": ">=8.0.0"
```
rwjblue: Before you publish, probably should bump to >= 10.0.0 (Node 8 is EOL)

packages/cli/yarn.lock
```diff
@@ -0,0 +1,2667 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
rwjblue: I think I assumed that we would only have one `yarn.lock` at the top level of the project.

.eslintrc
scalvert: Can do!

.eslintignore
scalvert: Yes.

package.json
scalvert: Fixed. This was an easy change, and a good one.

package.json
scalvert: Changed to just calling the yarn script. I wasn't sure this was possible based on the docs, but makes sense that it would be as it's just executing an arbitrary command.

packages/cli/package.json
scalvert: I think the generator did some funky concatenating. Fixed up.

packages/cli/package.json
scalvert: I think the generator did some funky concatenating. Fixed up.

packages/cli/README.md
scalvert: 🤷‍♂

packages/cli/package.json
scalvert: Updated to a proper range. Should probably update the oclif generator too.

packages/cli/yarn.lock
scalvert: Ya I believe this was generated by me running the oclif generator in a subdir of the `packages` dir. I've deleted the `yarn.lock`, and verified it doesn't regenerate when `yarn` is run from the root.

PR #3, opened 2020-02-11
Adding Checkup spec

scalvert:
[Rendered](https://github.com/checkupjs/checkup/blob/spec/SPEC.md)

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,0 +1,157 @@
+# Checkup
+
+> A health check for your project
+
+# Motivation
+
+Often, when building large (and ambitious) applications, the state of the code is almost in constant flux. This is a good thing, and is indicative of a healthy, maintained application. The dark side to this can be the app feeling like it's constantly in a half state - migrations from old APIs to new are taking place, patterns are being reinvented and rolled out, and tests are being created and maintained (and sometimes 😱 skipped). Keeping track of all these ongoing changes in your codebase can be challenging.
+
+Checkup aims to provide you with insights into your codebase. You can see things like dependencies, types used, test types used, etc. allowing you to have a more full, high level view of your codebase at any one point in time. It also allows consumers to author plugins, providing an extensibility mechanism to address varied needs. This will help you with making maintenance decisions, charting progress of migrations, and keeping up with the general health of your codebase.
+
+# Key Ideas
+
+## Provide Project Information
+
+Knowledge is power. Checkup wants to give you constant insights into the current and historical state of your codebase.
+
+## Provide Health Checkups
+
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -0,0 +1,157 @@
...
+
+It's likely that we'll want to use specific lint rules both in Checkup and configured to run in a project. Again, Checkup's role is one of a data aggregator, and lint rules provide more fine-grained, actionable guidance on changes within source code.
+
+### Results
+
+`Tasks` return `Results`, which are in turn formatted for output based on what data they return.
+
+### Output
+
+`Results` are aggregated from `Tasks`, formatted, and output depending on their target. Output can be any of `console`, `PDF`, or `HTML`.
```
lbdm44: Should we add `JSON` to the output list?

SPEC.md
```diff
@@ -0,0 +1,161 @@
...
+
+- Organization specific
+
+  Any plugins that are specific to an organization may not make sense to include as part of checkup directly. Ensuring that specific internal APIs are used the correct way, for instance, may be a concern best left to consuming organizations.
+
+- External library specific
+
+  Plugins that are specific to external libraries, such as `ember-concurrency` for example, may be best left as externally authored plugins that are ultimately configurable.
+
+## Configuration
```
mahirshah: I imagine each project would maintain a .checkuprc file or similar in order to configure which Tasks to run and the config options passed to those Tasks?

SPEC.md
scalvert: Good call. Ya I think we should add it.

SPEC.md
scalvert: Agree, which is what our original thought was. Either that or package.json (or both!).

I wanted to keep those kinds of details out of this spec though, as this is more the high level requirements.

SPEC.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Tasks
+
+`Tasks` are the main building block of Checkup. They should provide a thin, simple interface that allows consumers to author tasks in the most straightforward manner.
+
+Tasks fall into one of two categories:
+
+#### `JSON`
+
+A JSON task is used to read JSON files, whether project information files (such as `project.json`), or configuration files (such as `.eslinrc.json` and the like). These tasks simply gather information from disparate sources to aggregate into a report.
```
suchitadoshi1987: this is a totally valid example, but should we use the `package.json` as an example instead?

scalvert: Derp. Thanks for catching this!

scalvert:
@rwjblue @lbdm44 @suchitadoshi1987 @mahirshah 

If none of you have any comments as to the overall high-level strategy outlined in this spec, I'm going to merge it by EOD Friday. We can always amend/update it after - it's not set in stone.

lbdm44:
LGTM

PR #4, opened 2020-02-11
Adding core infra from external package

scalvert:
This PR merges the core files from `ember-cli-checkup` into the new checkupjs/checkup repo. 

⚠️ This does _not_ include task tests. This was intentional. Adding the existing tests would require adding a number of ember-specific packages to this repository. I wanted to keep those to a minimum, so this PR merely merges the core files in for further refactoring.


scalvert:
I'm going to merge this as is as it's a stepping stone on our way to revamping some of the architecture.

PR #25, opened 2020-02-20
Add a vs-code workspace

lbdm44:
Adding a vs-code workspace so that we can suggest plugins and expand to default settings/folders in the future.


scalvert:
LGTM. Thanks!

PR #26, opened 2020-02-21
feat(core): Add support for checkup config file

mahirshah:
Add support for a consumer provided checkup config file, such as a .checkuprc, loaded via cosmiconfig. Add scaffolding to pass the checkup config to the
checkup task runner.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -11,6 +11,7 @@ A CLI that provides health check information about your project.
 <!-- toc -->
 
 - [Usage](#usage)
+- [Configuration](#configuration)
 - [Commands](#commands)
   <!-- tocstop -->
 
@@ -34,6 +35,32 @@ USAGE
 ```
 
 <!-- usagestop -->
+
+# Configuration
+
+<!-- configuration -->
+
+checkup is designed to be completely configurable via a configuration object.
+
+checkup uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to find and load your configuration object. Starting from the current working directory, it looks for the following possible sources:
+                                        
+- a checkup property in package.json
+- a .checkuprc file
+- a checkup.config.js file exporting a JS object
+
+The search stops when one of these is found, and checkup uses that object. 
+
+The .checkuprc file (without extension) can be in JSON or YAML format. You can add a filename extension to help your text editor provide syntax checking and highlighting:
+
+- checkup.json
+- checkup.yaml / .checkup.yml
+- checkup.js
+
+The configuration object has the following properties:
+__
+<!-- TODO: Describe properties in CheckupConfig -->
+
+<!-- configurationstop -->
 
 # Commands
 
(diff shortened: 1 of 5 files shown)


packages/core/src/types/index.ts
```diff
@@ -18,3 +18,10 @@ export interface TaskItemData {
 }
 
 export type SearchPatterns = Record<string, string[]>;
+
+export interface TaskConfig {}
+
+export interface CheckupConfig {
+  // an object from task name to task configuration
+  tasks: Record<string, TaskConfig>;
```
lbdm44: I believe we want to have a config that defines where our project's plugins, rather than the individual tasks.

i.e. "@checkup/plugin-ember" and/or "@company/plugin-default-checkup-tasks"

mahirshah: Correct, I am thinking that this CheckupConfig will also have a "plugins: string[]" field, that lists the plugins that you want to use, although I think that is out of scope for this PR. I was imagining certain Tasks will also need some options passed in. For example, you can turn on/off a Task, set the severity, etc. Thoughts?

scalvert: Ya I think we want both, similar to what eslint provides. You should be able to:

1. Specify which plugins you want to load
1. Specify any individual task overrides you want to configure, which would allow you to modify the proposed `Urgency` property in the `TaskResult` (yet to be added). This would allow projects to individually configure checkup based on their needs, and temporarily upgrade/downgrade `Urgency` based on those needs.

packages/cli/README.md
```diff
@@ -35,6 +36,32 @@ USAGE
 
 <!-- usagestop -->
 
+# Configuration
+
+<!-- configuration -->
+
+checkup is designed to be completely configurable via a configuration object.
+
+checkup uses [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) to find and load your configuration object. Starting from the current working directory, it looks for the following possible sources:
```
scalvert: Nice choice!

packages/cli/src/index.ts
```diff
@@ -16,10 +17,10 @@ class CheckupCli extends Command {
 
   async run() {
     let { args, flags } = this.parse(CheckupCli);
-
     await this.config.runHook('register-tasks', {});
+    const config = await getConfig();
```
scalvert: Thought it's fine for now, I think we need to move this up so that it's one of the first things that gets called. The reason for this is I think we want to execute in this order:

1. Read config
1. Based on config, load configured plugins and tasks
1. Invoke `register-tasks` hook based on config
1. Run tasks. 

You don't necessarily need to change it now, but ultimately we'll need to.

mahirshah: Sure, after rebasing, I removed this anyway. I will work on the plugin configuration/loading in the next PR.

scalvert: Thanks!

scalvert:
Thanks for this. This is a great start!

PR #30, opened 2020-02-23
feat(core): Add Support for Plugins via Config File

mahirshah:
Add support for dynamically loading plugins specified in the consumer
checkup config plugins field.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -69,11 +69,9 @@ $ yarn add -D @checkup/plugin-ember
 To configure plugins, use the plugins key in your configuration file, which contains a list of plugin names.
 ```json
 {
-  // ...
   "plugins": [
     "@checkup/plugin-ember"
   ]
-  // ... 
 }
 ```
 <!-- TODO: Describe properties in CheckupConfig -->
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -57,7 +57,25 @@ The .checkuprc file (without extension) can be in JSON or YAML format. You can a
 - checkup.js
 
 The configuration object has the following properties:
-__
+
+## plugins
+
+checkup supports the use of plugins. Before using the plugin, you have to install it using npm/yarn.
+
+```sh-session
+$ yarn add -D @checkup/plugin-ember
+```
+
+To configure plugins, use the plugins key in your configuration file, which contains a list of plugin names.
+```json
+{
+  // ...
+  "plugins": [
+    "@checkup/plugin-ember"
+  ]
+  // ... 
+}
+```
 <!-- TODO: Describe properties in CheckupConfig -->
 
 <!-- configurationstop -->
--- a/packages/cli/__tests__/index-test.ts
+++ b/packages/cli/__tests__/index-test.ts
@@ -1,16 +1,33 @@
 import { stdout } from '@checkup/test-helpers';
-
+import Project = require('fixturify-project');
 import cmd = require('../src');
+const FixturifyProject = require('fixturify-project');
 
 describe('@checkup/cli', () => {
+  let project: Project;
+
+  beforeEach(function() {
+    project = new FixturifyProject('checkup-project', '0.0.0');
+    project.addDevDependency('@checkup/plugin-ember', '0.0.0');
+    project.files['.checkuprc'] = JSON.stringify({
+      plugins: ['@checkup/plugin-ember'],
+      tasks: [],
+    });
+    project.writeSync();
+  });
+
+  afterEach(function() {
+    project.dispose();
+  });
+
   it('should output checkup result', async () => {
-    await cmd.run([]);
+    await cmd.run([project.baseDir]);
 
     expect(stdout()).toMatchSnapshot();
   });
 
   it('should output checkup result in JSON', async () => {
-    await cmd.run(['--json']);
+    await cmd.run(['--json', project.baseDir]);
 
     expect(stdout()).toMatchSnapshot();
   });
--- a/packages/cli/__tests__/index-test.ts
+++ b/packages/cli/__tests__/index-test.ts
@@ -2,7 +2,7 @@ import { stdout, CheckupProject, Plugin } from '@checkup/test-helpers';
 import cmd = require('../src');
 
 describe('@checkup/cli', () => {
-  describe('base cases', () => {
+  describe('normal cli output with plugins', () => {
     let project: CheckupProject;
 
     beforeEach(function() {
@@ -60,9 +60,15 @@ describe('@checkup/cli', () => {
 
       expect(stdout()).toMatchSnapshot();
     });
+
+    it('should run a single task if the task option is specified', async () => {
+      await cmd.run(['--task', 'mockTask', project.baseDir]);
+
+      expect(stdout()).toMatchSnapshot();
+    });
   });
 
-  describe('error cases', () => {
+  describe('cli error cases', () => {
     it('should error if no checkup config is present', async () => {
       const project = new CheckupProject('checkup-project', '0.0.0');
       project.writeSync();
--- a/packages/cli/__tests__/index-test.ts
+++ b/packages/cli/__tests__/index-test.ts
@@ -67,7 +67,9 @@ describe('@checkup/cli', () => {
       const project = new CheckupProject('checkup-project', '0.0.0');
       project.writeSync();
 
-      await expect(cmd.run([project.baseDir])).rejects.toThrowErrorMatchingSnapshot();
+      await expect(cmd.run([project.baseDir])).rejects.toThrowErrorMatchingInlineSnapshot(
+        `"Could not find a checkup configuration starting from the given path: ${project.baseDir}. See https://github.com/checkupjs/checkup/tree/master/packages/cli#configuration for more info on how to setup a configuration."`
+      );
 
       project.dispose();
     });
@@ -79,7 +81,9 @@ describe('@checkup/cli', () => {
       });
       project.writeSync();
 
-      await expect(cmd.run([project.baseDir])).rejects.toThrowErrorMatchingSnapshot();
+      await expect(cmd.run([project.baseDir])).rejects.toThrowErrorMatchingInlineSnapshot(
+        `"Cannot find module '@checkup/unknown-plugin' from '${project.baseDir}'"`
+      );
 
       project.dispose();
     });
--- a/packages/cli/__tests__/index-test.ts
+++ b/packages/cli/__tests__/index-test.ts
@@ -1,18 +1,46 @@
-import { stdout } from '@checkup/test-helpers';
-import Project = require('fixturify-project');
+import { stdout, CheckupProject, Plugin } from '@checkup/test-helpers';
 import cmd = require('../src');
-const FixturifyProject = require('fixturify-project');
 
 describe('@checkup/cli', () => {
-  let project: Project;
+  let project: CheckupProject;
 
   beforeEach(function() {
-    project = new FixturifyProject('checkup-project', '0.0.0');
-    project.addDevDependency('@checkup/plugin-ember', '0.0.0');
-    project.files['.checkuprc'] = JSON.stringify({
-      plugins: ['@checkup/plugin-ember'],
-      tasks: [],
-    });
+    const plugin = new Plugin.PluginBuilder('@checkup/plugin-mock')
+      .addTask('mockTask', {
+        async run() {
+          return Promise.resolve({
+            toJson() {
+              return {
+                mockTask: 5,
+              };
+            },
+            toConsole() {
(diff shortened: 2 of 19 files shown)


packages/cli/README.md
```diff
@@ -57,7 +57,25 @@ The .checkuprc file (without extension) can be in JSON or YAML format. You can a
...
+
+## plugins
+
+checkup supports the use of plugins. Before using the plugin, you have to install it using npm/yarn.
+
+```sh-session
+$ yarn add -D @checkup/plugin-ember
+```
+
+To configure plugins, use the plugins key in your configuration file, which contains a list of plugin names.
```
scalvert:
```suggestion
To configure plugins, use the plugins key in your .checkuprc file, which contains a list of plugin names.
```

packages/cli/README.md
```diff
@@ -57,7 +57,25 @@ The .checkuprc file (without extension) can be in JSON or YAML format. You can a
...
 
 The configuration object has the following properties:
-__
+
+## plugins
+
+checkup supports the use of plugins. Before using the plugin, you have to install it using npm/yarn.
+
+```sh-session
+$ yarn add -D @checkup/plugin-ember
```
scalvert: We're going to need to figure out how this is going to work with checkup being a globally installed package vs. a locally installed one. Essentially we need to be able to potentially dynamically load packages from the registry for any plugins, rather than forcing users to install a set of packages to make checkup work.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
+import checkupCore = require('@checkup/core');
 
 describe('@checkup/cli', () => {
+  const defaultConfig: checkupCore.CheckupConfig = {
+    plugins: ['@checkup/plugin-ember'],
+    tasks: {},
+  };
+
   it('should output checkup result', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(Promise.resolve(defaultConfig));
```
scalvert: So I want to stay away from these kinds of tests. I'm not hugely in favor of using mocks generally, as I believe they're too hard to maintain. 

I'd prefer we rewrite this using `fixturify-project`, which allows you to dynamically create fixture data that lives on the real filesystem in a tmp directory, and is subsequently cleaned up after the tests complete. We can write a directory _with_ a real `.checkuprc` file, which will allow us to actually test the real functionality. 

To use `fixturify-project`: https://github.com/checkupjs/checkup/blob/f2c372f0817f1d736dcdb741d4f2c08e8c726892/packages/plugin-ember/__tests__/project-info-task-test.ts#L13-L19

To write a real test that can use the tmp directory generated, pass `fixturifyProject.baseDir` as the `path` argument passed to checkup's CLI. That will then run checkup in the generated directory, and it will actually test the real implementation.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
     await cmd.run([]);
 
     expect(stdout()).toMatchSnapshot();
   });
 
   it('should output checkup result in JSON', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(Promise.resolve(defaultConfig));
```
scalvert: Same as above.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
     await cmd.run(['--json']);
 
     expect(stdout()).toMatchSnapshot();
   });
 
   it('should output checkup result for project info task', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(Promise.resolve(defaultConfig));
```
scalvert: Same as above.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
     await cmd.run(['--task', 'ProjectInfo']);
 
     expect(stdout()).toMatchSnapshot();
   });
 
   it('should output checkup result for dependencies task', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(Promise.resolve(defaultConfig));
```
scalvert: Same as above.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
     await cmd.run(['--task', 'Dependencies']);
 
     expect(stdout()).toMatchSnapshot();
   });
 
   it('should output checkup result for types task', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(Promise.resolve(defaultConfig));
```
scalvert: Same as above.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,33 +1,55 @@
...
+    await cmd.run(['--task', 'Types']);
+
+    expect(stdout()).toMatchSnapshot();
+  });
+
+  it('should output nothing if no plugins are present', async () => {
+    checkupCore.getConfig = jest.fn().mockReturnValue(
```
scalvert: Same as above.

packages/cli/package.json
```diff
@@ -59,9 +59,7 @@
   "main": "lib/index.js",
   "oclif": {
     "bin": "checkup",
-    "plugins": [
-      "@checkup/plugin-ember"
-    ]
+    "plugins": []
```
scalvert: I think this can just be removed.

packages/core/src/types/index.ts
```diff
@@ -22,6 +22,8 @@ export type SearchPatterns = Record<string, string[]>;
 export interface TaskConfig {}
 
 export interface CheckupConfig {
+  // a list of plugin names that should be loaded for checkup
+  plugins: string[];
   // an object from task name to task configuration
   tasks: Record<string, TaskConfig>;
 }
```
scalvert: Can we just use jsdoc comments instead for this interface? (All of them should be documented, as then they're documented in the output `d.ts` file.

packages/cli/README.md
mahirshah: I'm hesitant to say `.checkuprc` file here as cosmiconfig allows for different types of configuration (checkup key in package.json or checkup.config.js for example). Thoughts?

packages/cli/README.md
mahirshah: Good call out, I haven't thought about how this will work with globally installed checkup yet. If I understand correctly, we should be consistent with how eslint handles plugins? i.e. if you are running checkup via the global installation look for the global installation of the plugin. If its being run via the local installation look for the local installation of the plugin. Let me know if this is what you were thinking

packages/cli/README.md
scalvert: So do we want to support multiple mechanisms for loading config, or one? If we support multiple, I could imagine supporting the package.json and a file, but I'd want to recommend a specific file name to keep it simple.

One of the reasons I'd rather have it be simply a config file is that we no longer have to refer to `oclif` as part of configuration, as that could end up being confusing for consumers who don't even know what oclif is. The config in the package.json would require this, unless we wrote our own config using a custom key (`checkup`). If we simply direct them to use a config file all the time, we don't have this problem. 

What do you think?

packages/cli/README.md
scalvert: Ya this is sorta what I'm thinking. We should discuss this a bit more, but for now you could force an install on that directory by `cd`ing into it and using `execa` to `yarn install`.

packages/core/src/loaders/plugin-loader.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Given a list of oclif plugin names, attempt to resolve the plugins and load
+ * them. Note: in order to use the plugins during an oclif command, the loaded
+ * plugins must be pushed onto the oclif config plugins array.
+ * @param pluginNames - plugin names to load
+ * @param resolutionBaseDir - the base directory to resolve plugins from
+ * @return Promise containing the loaded plugins
+ */
+export async function loadPlugins(pluginNames: string[], resolutionBaseDir: string) {
+  const plugins = pluginNames
+    .map(pluginName => require.resolve(pluginName, { paths: [resolutionBaseDir] }))
```
rwjblue: Let's swap this to using [resolve](https://www.npmjs.com/package/resolve), would look something like:

```suggestion
    .map(pluginName => resolve.sync(pluginName, { basedir: resolutionBaseDir }))
```

packages/cli/README.md
mahirshah: resolved offline. Plugins will be resolved starting from config path passed into the checkup command

packages/cli/README.md
scalvert: I'm just going to resolve this for now. We can bikeshed this another time :)

packages/cli/__tests__/index-test.ts
```diff
@@ -1,17 +1,91 @@
-import { stdout } from '@checkup/test-helpers';
-
+import { stdout, CheckupProject, Plugin } from '@checkup/test-helpers';
 import cmd = require('../src');
 
 describe('@checkup/cli', () => {
-  it('should output checkup result', async () => {
-    await cmd.run([]);
+  describe('base cases', () => {
```
scalvert: Sorry to nit pick, but can we provide a more descriptive title for these groupings? It's be great for people to be able to tell immediately what is under test without having to parse the code below.

packages/cli/__tests__/index-test.ts
```diff
@@ -1,17 +1,91 @@
...
+    it('should output checkup result in JSON', async () => {
+      await cmd.run(['--json', project.baseDir]);
+
+      expect(stdout()).toMatchSnapshot();
+    });
   });
 
-  it('should output checkup result in JSON', async () => {
-    await cmd.run(['--json']);
+  describe('error cases', () => {
```
scalvert: Can you also add a more descriptive title here?

packages/cli/__tests__/index-test.ts
```diff
@@ -1,17 +1,91 @@
...
+    });
+
+    it('should output checkup result', async () => {
+      await cmd.run([project.baseDir]);
+
+      expect(stdout()).toMatchSnapshot();
+    });
 
-    expect(stdout()).toMatchSnapshot();
+    it('should output checkup result in JSON', async () => {
```
scalvert: Can you add a case that tests the `--task` flag?

packages/core/src/loaders/plugin-loader.ts
```diff
@@ -0,0 +1,20 @@
+import * as Config from '@oclif/config';
+import * as resolve from 'resolve';
+
+/**
+ * Given a list of oclif plugin names, attempt to resolve the plugins and load
+ * them. Note: in order to use the plugins during an oclif command, the loaded
+ * plugins must be pushed onto the oclif config plugins array.
+ * @param pluginNames - plugin names to load
+ * @param resolutionBaseDir - the base directory to resolve plugins from
+ * @return Promise containing the loaded plugins
```
scalvert: Can we add types for these to the jsdoc? This really helps with hints in the editor for JS consumers.

packages/core/src/utils/get-config.ts
```diff
@@ -1,18 +1,18 @@
 import { cosmiconfig } from 'cosmiconfig';
 import { CheckupConfig } from '../types';
 
-let config: CheckupConfig;
-const DEFAULT_CONFIG: CheckupConfig = { tasks: {} };
-
 /**
  * Get the checkup config via {@link cosmiconfig#search}
  * @return the parsed config file, if found, else the default config
```
scalvert: Can you add a type to this jsdoc?

packages/test-helpers/src/checkup-fixturify-project.ts
```diff
@@ -0,0 +1,33 @@
+'use strict';
+import { CheckupConfig } from '@checkup/core';
+import Plugin from './plugin';
+import FixturifyProject from 'fixturify-project';
+
+/**
+ * An extension of {@link Project} that adds methods specific to creating
+ * mock checkup projects.
+ */
+export default class CheckupFixturifyProject extends FixturifyProject {
```
scalvert: One thing I'm not sure of is how this class will interact with the `ember-cli-fixturify-project` class we have already. I think perhaps that class should extend from this, so if people are testing ember projects they can still get this core checkup project generation functionality too. Thoughts?

mahirshah: Yeah, I think `ember-cli-fixturify-project` should probably be moved to the plugin-ember package and extend off of this class. I imagine other plugins like `plugin-react` may create their own `react-fixturify-project` and extend off of this as well.

scalvert: 👍 we can do that in a separate PR.

PR #39, opened 2020-02-24
internal(infra): Utilize workspaces run for build commands

lbdm44:
Utilize `yarn workspaces run` for build command so we don't need to keep on updating our build command as our workspaces grow.

Create `clean` commands for each package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,11 +7,14 @@
   "license": "MIT",
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/core",
+    "packages/test-helpers",
+    "packages/plugin-ember",
+    "packages/cli"
   ],
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,11 +7,14 @@
   "license": "MIT",
   "private": true,
   "workspaces": [
-    "packages/*"
+    "packages/core",
+    "packages/test-helpers",
+    "packages/plugin-ember",
+    "packages/cli"
```
lbdm44: We need to specify the order here due to yarnpkg/yarn#7150

scalvert:
Oh HA! I didn't see this. I just opened https://github.com/checkupjs/checkup/pull/42. Which do you think we should go with? I'm leaning towards `wsrun`, as it works in parallel.

lbdm44:
Duplicate of #42

PR #42, opened 2020-02-25
Adding wsrun package to execute tasks in dependency order and parallel

scalvert:
This changes the top level build command to use `wsrun` instead of an unmanageable script config.

To run from the top level:

`yarn wsrun --stages build` will build all packages, in stages, starting from those that don't depend on other packages.
`yarn wsrun watch` will run `yarn watch` on every individual package, in parallel.
`yarn wsrun test` will run `yarn test` on every individual package, in parallel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "yarn workspace @checkup/core build && yarn workspace @checkup/test-helpers build && yarn workspace @checkup/plugin-ember build && yarn workspace @checkup/cli build",
-    "build:watch": "yarn workspace @checkup/core build -w && yarn workspace @checkup/plugin-ember build -w && yarn workspace @checkup/cli build -w",
+    "build": "echo 'Please use `yarn wsrun --stages build` to build all workspace packages in dependency order.'",
+    "build:watch": "echo 'Please use `yarn wsrun watch` to run the build in watch mode for each individual package.'",
     "test": "yarn workspaces run test",
     "lint": "eslint . --cache --ext .ts"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,8 +10,8 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "yarn workspace @checkup/core build && yarn workspace @checkup/test-helpers build && yarn workspace @checkup/plugin-ember build && yarn workspace @checkup/cli build",
-    "build:watch": "yarn workspace @checkup/core build -w && yarn workspace @checkup/plugin-ember build -w && yarn workspace @checkup/cli build -w",
+    "build": "echo 'Please use `yarn wsrun --stages build` to build all workspace packages in dependency order.'",
```
lbdm44: We need to either update the workflow to use `wsrun` or just have the `build` command call out to `wsrun` itself

package.json
```diff
@@ -10,8 +10,8 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "yarn workspace @checkup/core build && yarn workspace @checkup/test-helpers build && yarn workspace @checkup/plugin-ember build && yarn workspace @checkup/cli build",
-    "build:watch": "yarn workspace @checkup/core build -w && yarn workspace @checkup/plugin-ember build -w && yarn workspace @checkup/cli build -w",
+    "build": "yarn wsrun --stages build",
+    "build:watch": "yarn wsrun watch",
     "test": "yarn workspaces run test",
```
lbdm44: Should we also use `wsrun` for tests? To be consistent with the other cross cutting package scripts

scalvert: Ya let's do that.

PR #49, opened 2020-02-28
fix(internal): Force task ordering

lbdm44:


lbdm44:
this was an example PR, closing in favor of #57

PR #53, opened 2020-02-28
Add Validation to Checkup Config

mahirshah:
Use io-ts to wrap CheckupConfig types so they can be accessed during
runtime and used to validate the configuration found via cosmiconfig

### DIFF
--- a/packages/cli/__tests__/__snapshots__/index-test.ts.snap
+++ b/packages/cli/__tests__/__snapshots__/index-test.ts.snap
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`@checkup/cli cli error cases should error if the config is malformed 1`] = `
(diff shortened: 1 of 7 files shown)


packages/cli/__tests__/__snapshots__/index-test.ts.snap
```diff
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`@checkup/cli cli error cases should error if the config is malformed 1`] = `
+"Checkup configuration is malformed:
+For key plugins expected type Array<string>, but got undefined
+For key tasks.someTask.isEnabled expected type boolean, but got \\"foo\\""
```
scalvert: Should we just drop the `For key` part?

```suggestion
`plugins` expected type Array<string>, but got undefined
`tasks.someTask.isEnabled` expected type boolean, but got \\"foo\\""
```

PR #57, opened 2020-03-02
WIP: Task prioritization

scalvert:
This PR starts to tackle a way to provide prioritization for tasks. The first step is to refactor the task registration system to remove the unnecessary layer of indirection via `registerTask` -> `TaskList.addTask`. Now, the `TaskList` class directly registers the task without the intermediary of the `registerTask` function provided by the `tasks` module. That module has been removed.

Also updates the `PluginBuilder` for tests to utilize this new registration system.

TODO:

- Extend `TaskList` to provide multiple `Maps` that correspond to categories
- Allow for `PriorityQueue`s within those category `Maps` to allow for sorting of  tasks based on priority


packages/plugin-ember/src/hooks/register-tasks.ts
```diff
@@ -3,10 +3,10 @@ import { DependenciesTask, EmberProjectTask, TypesTask } from '../tasks';
...
 
 // TODO: Determine correct type for options
-const hook: Hook<'register-tasks'> = async function({ registerTask }: any) {
-  registerTask(EmberProjectTask.taskName, EmberProjectTask);
-  registerTask(DependenciesTask.taskName, DependenciesTask);
-  registerTask(TypesTask.taskName, TypesTask);
+const hook: Hook<'register-tasks'> = async function({ cliArguments, tasks }: any) {
+  tasks.registerTask(EmberProjectTask.taskName, new EmberProjectTask(cliArguments));
+  tasks.registerTask(DependenciesTask.taskName, new DependenciesTask(cliArguments));
+  tasks.registerTask(TypesTask.taskName, new TypesTask(cliArguments));
```
lbdm44: Was going to do this today, but looks like you could probably just do it as part of this refactor.

We can just pass in the instance and add `taskName` to the interface of `Task` so that all tasks must define one. This allows us to then have `bucket` and `bucketPriority` (or something similar) to `Task` so they can be referenced when we are sorting the run/output order.

```suggestion
  tasks.registerTask(new EmberProjectTask(cliArguments));
  tasks.registerTask(new DependenciesTask(cliArguments));
  tasks.registerTask(new TypesTask(cliArguments));
```

scalvert: Ya I thought of this, and this particular code is taking shape as I go along. One thing is that this value is constant - it doesn't really change for each instance (though it's a moot point, as we only ever have one instance). Making it static did seem correct, but I _do_ agree with the ability to enforce implementers to implement these properties (though we can't really enforce they've initialized them correctly except via validation, which could be done right here as tasks are added).

Let me finalize some of the things I'm doing right now and I'll move ahead with this next part.

packages/cli/src/priority-map.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  *valuesIterator() {
+    for (let [, map] of this.maps) {
+      for (let [, task] of map) {
+        yield task;
+      }
+    }
+  }
+}
```
lbdm44: We also need to ensure ordering of the tasks within each "priority" via alphabetical ordering, otherwise we are going to fall into the same problem with potential random ordering within each priority's map

packages/cli/src/task-list.ts
```diff
@@ -1,66 +1,94 @@
...
 
   /**
-   * @method addDefault
+   * @method registerTask
    *
    * Adds a default task to the task list, which is executed as part of checkup.
-   *
-   * @param taskConstructor {TaskConstructor} a constructor representing a Task class
+   * @param taskName {TaskName}
+   * @param task {Task}
```
lbdm44: Given that it is hard to keep up to date with a shifting pre-alpha codebase, should we just remove JSDoc until we attempt to publish a public version of checkup? (i.e. this JSDoc isnt eve up to date 😉)

packages/cli/src/task-list.ts
```diff
@@ -1,66 +1,94 @@
...
    */
-  addTask(taskConstructor: TaskConstructor, cliArguments: any) {
-    this.defaultTasks.push(new taskConstructor(cliArguments));
+  registerTask(taskName: TaskName, task: Task, taskClassification: TaskClassification) {
```
lbdm44: I recommend adding a default classification that way implementors can modify if they dont want the default.

```suggestion
registerTask(
  taskName: TaskName,
  task: Task,
  taskClassification: TaskClassification = {
    category: Category.Insights
    priority: Priority.Medium
  }
) {
```
see comment about `Insights`

packages/plugin-ember/src/hooks/register-tasks.ts
lbdm44: As per conversation, we will tackle this at a later point in time.

packages/core/src/types/index.ts
```diff
@@ -14,16 +12,27 @@ export interface Parser {
   execute(paths: string[]): JsonObject;
 }
 
+export const enum Category {
+  Core,
+  Migration,
```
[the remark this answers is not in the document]
scalvert: Sure. We can add that.

packages/cli/src/priority-map.ts
scalvert: So right now it's insertion order. I'd prefer to do that over alpha, which is very arbitrary. 

Insertion order doesn't really help us with async hooks running, but it just feels odd to alpha sort, which doesn't really mean much. What do you think?

packages/cli/src/task-list.ts
scalvert: I'm sure there's an eslint plugin that can look at jsdoc comments. We should likely just add that.

packages/cli/src/task-list.ts
scalvert: I think you'd _always_ want to define the classification. I would actually say we should enforce this, as it doesn't make sense to have a default, as it essentially has no meaning at that point. Thoughts?

packages/cli/src/priority-map.ts
lbdm44: I agree with alpha being odd, but we need it to be deterministic, because insertion order in testing isnt guaranteed. I.e. in my plugin im working on right now, each time i run tests i need to re-create the snapshot of the checkup/cli tests since it changes output location every run.

We could change our data-structure when storing to solve this as well, tracking insertion order in an array and a map from task-name to index of the insertion array.

packages/core/src/types/index.ts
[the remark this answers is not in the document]
scalvert: Done!

packages/cli/src/priority-map.ts
lbdm44: Or rather, the order of insertion is not guaranteed in tests because of how hooks are working.

scalvert: Can you help me understand how order is not guaranteed in tests currently, when testing a single package in the CLI tests? We're creating a dummy plugin that explicitly registers tasks in a specific order, dictated by the hook. How is it that it's changing each time?

lbdm44: Yes, I can walk you though what i was seeing, probably easier to pair on that, but if you think about this at a high level, we are doing the exact same thing now for each priority that we were doing previously for all tasks. If we had a problem previously, we are going to have that same problem at the p-queue level if there are multiple tasks defined for those p-queues

packages/cli/src/task-list.ts
scalvert: Added one. :)

packages/cli/src/priority-map.ts
lbdm44: This "problem" may have been solved with the change that @mahirshah made for loading in plugins, since now `@checkup/cli` doesn't define any plugins and the tests are now using the fixturify plugin builder.

scalvert: Unsure. Basically the maps operate on insertion order, the categories are in a logical sequence (`Core`, `Migrations`, `Insights`), which would iterate in order. The next level is the `PriorityMap`, which is always in insertion order of `High`, `Medium`, `Low`. I do believe that we'd have issues _inside_ each of these priority maps, but in tests we don't see this right now as we're not creating multiple competing plugins with multiple tasks (we should add those to expose this more!). So, I supposed what I'm trying to understand is how you're seeing order changing in light of us not having the above in our tests.

scalvert: I think what you're saying has some merit, and I think we should for sure look into this in subsequent PRs. I'd like to just move ahead with this one now since it's already grown a bit. Are you OK with us leaving this for now?

packages/cli/src/task-list.ts
lbdm44: I don't think it has no meaning if you don't define it, it just means that you are ok with wherever the framework places your output.

I don't feel very strongly about this, if you are on the firming on the side of no defaults, then we can go that way.

On another note, as I review this I think we are going to need to do a pass over our code once we get closer to launch and throw errors in places where crucial data is not provided, we cannot guarantee our customers are using TS and having better error messages than "x is undefined" will help out a ton.

packages/cli/src/priority-map.ts
lbdm44: Yeah, when i get free time ill create a PR that shows this error case and we can figure out how to address it then.

packages/cli/__tests__/task-list-test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+describe('TaskList', () => {
+  it('can create an instance of a TaskList', () => {
+    let taskList = new TaskList();
+
+    expect(taskList).toBeInstanceOf(TaskList);
+    expect(taskList.categories.size).toEqual(3);
+  });
+
+  it('registoryTask adds a task to the TaskList', () => {
```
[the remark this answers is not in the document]
scalvert: Good catch!

PR #60, opened 2020-03-02
Refactor Checkup Configuration Loading

mahirshah:
Refactor how checkup configurations are loaded internally:
Create a CheckupConfigService that can be used to load a checkup config
via a loader function, map a config, and write the config back to the
loaded file if needed.
Expose a cosmiconfig loader factory to create a CheckupConfigLoader that
relies on cosmiconfig to find the config file.

Why?
===
Future PRs will allow for:
1) Getting a config from a given filepath via the --config option
2) Adding a config:init command to initialize a config

This PR allows us to easily implement the above 2 features by simply implementing a custom loader and any required CheckupConfigMappers

For example, imagine a initializing a config:
```
await CheckupConfigService
  .load(() => { config: ..., ... })
  .map(addPlugin('@checkup/plugin-default'))
  .write();
```


scalvert:
Would you mind adding a little prose around the _why_ behind this change? What does this added complexity allow us to do?

packages/core/src/configuration/loaders/cosmiconfig-loader-factory.ts
```diff
@@ -0,0 +1,23 @@
+import { CheckupConfigLoader } from '../../types';
+import CosmiconfigService from '../cosmiconfig-service';
+
+/**
+ * A factory function to get an instance of a {@link CheckupConfigLoader} that
+ * uses {@link cosmiconfig#search} to get a {@link CheckupConfig} object
+ * @param {string} basePath - the base path to start the config search in
+ */
+const cosmiConfigLoaderFactory: (basePath: string) => CheckupConfigLoader = (
```
scalvert: How is this a factory exactly?

packages/cli/src/index.ts
```diff
@@ -49,7 +50,8 @@ class Checkup extends Command {
   async run() {
     let { args, flags } = this.parse(Checkup);
     let registeredTasks: Map<TaskName, TaskConstructor>;
-    const checkupConfig = await getConfig(args.path);
+    const configService = await CheckupConfigService.load(CosmiconfigLoaderFactory(args.path));
```
scalvert: Can't we just abstract this into the `CheckupConfigService`? Why do we need this factory too?

packages/core/src/configuration/loaders/cosmiconfig-loader-factory.ts
mahirshah: Its a function that creates an instance of a CheckupConfigLoader

packages/cli/src/index.ts
mahirshah: abstract which part into `CheckupConfigService`? We want consumers to be able to inject their own loader into the CheckupConfigService in order to allow for the use cases described above (it also keeps CheckupConfigService lean + easy to test)

packages/core/src/configuration/loaders/cosmiconfig-loader-factory.ts
scalvert: So I usually think of factories as creators of instances that implement an interface, but are different types. Effectively, facilitating object creation for polymorphism. [ref](https://www.dofactory.com/net/factory-method-design-pattern).

You just need a simple `getConfigLoader` and return that instance, no?

PR #81, opened 2020-03-04
fix(plugin mocker): Cleans up the plugin mocker to ensure type of tasks are validated

scalvert:


packages/core/package.json
```diff
@@ -6,9 +6,8 @@
   "bugs": "https://github.com/checkupjs/checkup/issues",
   "dependencies": {
     "fp-ts": "^2.5.3",
-    "io-ts": "^2.1.2",
+    "io-ts": "^2.1.2"
     "js-yaml": "^3.13.1",
-    "resolve": "^1.15.1"
```
mahirshah: why was this dep removed?

scalvert: Adding it back. Bad merge.

PR #83, opened 2020-03-05
feat(plugin-ember-octane): Create plugin-ember-octane

lbdm44:
This is not fully functional, since there is a lot of moving parts, we are checking this in and refactoring with smaller PRs going forward.


packages/plugin-ember-octane/src/tasks/octane-migration-status-task.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.esLintEngine = new CLIEngine({
+      parser: 'babel-eslint',
+      parserOptions: {
+        ecmaVersion: 2018,
+        sourceType: 'module',
+        ecmaFeatures: {
+          legacyDecorators: true,
+        },
+      },
+      plugins: ['ember'],
```
scalvert: This seems very sensible to me, in that we can tie the specific ruleset directly with the task. I'm 👍 on this.

PR #87, opened 2020-03-06
PDF reporter

scalvert:
This PR gets the basic structure for the PDF reporter working. It changes a few things:

1. The `--json` flag has been replaced with `--reporter`, with the options `stdout`, `json`, or `pdf`
2. The JSON output from a result is more constrained now, and contains both the task's `meta` and a `result` property.
3. The JSON results are aggregated now by `category`, then by `priority`. 

TODO:

- [x] Tests

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,14 +5,13 @@
     "ecmaVersion": 2017,
     "sourceType": "module"
   },
-  "plugins": ["@typescript-eslint", "filenames", "jest", "jsdoc", "prettier"],
+  "plugins": ["@typescript-eslint", "filenames", "jest", "prettier"],
   "extends": [
     "eslint:recommended",
     "prettier",
     "prettier/@typescript-eslint",
     "plugin:jest/recommended",
     "plugin:jest/style",
-    "plugin:jsdoc/recommended",
     "plugin:unicorn/recommended"
   ],
   "env": {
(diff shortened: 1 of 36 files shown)


.eslintrc
```diff
@@ -5,14 +5,13 @@
     "ecmaVersion": 2017,
     "sourceType": "module"
   },
-  "plugins": ["@typescript-eslint", "filenames", "jest", "jsdoc", "prettier"],
+  "plugins": ["@typescript-eslint", "filenames", "jest", "prettier"],
```
scalvert: Sorry, this was bugging me.

.eslintrc
```diff
@@ -24,7 +23,8 @@
     "prettier/prettier": "error",
     "no-unused-vars": "off",
     "@typescript-eslint/no-unused-vars": "error",
-    "no-global-assign": ["error", { "exceptions": ["console"] }]
+    "no-global-assign": ["error", { "exceptions": ["console"] }],
+    "unicorn/prevent-abbreviations": "off"
```
scalvert: So was this.

.eslintrc
mahirshah: thank you!

mahirshah:
Overall looks good, although I was originally thinking that the reporters would actually each be a diff package, each being an oclif plugin that has a hook like registerReporter. The reporters would have a simply export a report method that takes in the task results. Thus the consumer of checkup would define in their `.checkuprc` something like `reporters: ['@checkup/pdf-reporter', '@foo/docx-reporter']`. This would let you have multiple reporters and keep your reporting consistent over multiple runs of the cli (which you can override via a cli option if necessary). Thoughts?

scalvert:
That seems fine, but I think prematurely complicated. I'd rather get something working, and if we need to extract it we do it then. I almost think the parsers stuff is prematurely complicated right now...

PR #97, opened 2020-03-09
Adding PDF function to task results

carakessler:
Working on PR that adds PDF function to all results to format as pdf. This eventually wont be void, but to minimize changes in subsequent PR, adding void function to each result


packages/core/src/types/index.ts
```diff
@@ -45,6 +45,7 @@ export type JsonTaskResult = {
 export interface TaskResult {
   stdout: () => void;
   json: () => JsonTaskResult;
+  pdf: () => void;
```
scalvert: Ultimately we'll be returning some sort of data structure that we'll use on the Handlebars side, correct?

PR #103, opened 2020-03-10
Add Config Init Command

mahirshah:
Add command to initialize a checkuprc file via `checkup config:init`.
Move to a true multi command cli, by moving main checkup command to the
`run` command


packages/core/src/configuration/config-init-service.ts
```diff
@@ -0,0 +1,68 @@
+import CheckupConfigService from './checkup-config-service';
+import { CheckupConfig, CheckupConfigFormat, ConfigMapper } from '../types';
+import addPlugin from './mappers/add-plugin';
+import getInitializationConfigLoader from './loaders/get-initialization-loader';
+import CosmiconfigService from './cosmiconfig-service';
+
+type QuestionNames = typeof ConfigInitService.questions[number]['name'];
+type Answers = Record<QuestionNames, string>;
+
+export default class ConfigInitService {
```
scalvert: I would much prefer this class was included in a directory that implied its intent.

`packages/core/src/generators/config/init.ts`

This is feeling very java-y :)

If we keep all our generators in that structure, you can organize and find them very easily on disc.

```
// generators/config/init.ts
checkup config:init 
```

```
// generators/task.ts
checkup task <taskName>
```

```
// generators/plugin.ts
checkup plugin <pluginName>
```

Inside the `generators` directory we can have a `generators/templates` directory that points to the files on disc. I'd really like to just use yoman here to create and invoke templates, similar to what oclif does. It's very powerful, and comes with all the inquirer, etc stuff you need.

packages/core/src/configuration/mappers/add-plugin.ts
```diff
@@ -0,0 +1,8 @@
+import { ConfigMapper } from '../../types';
+
+const addPlugin: (pluginName: string) => ConfigMapper = (pluginName: string) => config => {
```
scalvert: Why does this need to be in an external module? Can it not be inlined in the generator?

packages/core/src/configuration/config-init-service.ts
scalvert: Example: https://github.com/oclif/oclif/blob/master/src/generators/command.ts#L10

packages/cli/src/commands/config/init.ts
```diff
@@ -0,0 +1,29 @@
+import { Command } from '@oclif/command';
```
scalvert: Thoughts on making a top level command called `generate`, a la Ember?

```
checkup generate config
checkup generate task <taskName>
checkup generate plugin <pluginName>
```

mahirshah: hmm I think generate makes sense for `task` and `plugin`, but for config it seems like `config:init` or similar is better. Thoughts? You should only be generating a config once per project I imagine so to me it seems like a special case.

scalvert: But they're all code generators, no? I could see that argument, and I'm not super opinionated either way about this. We can always change it later if we want.

packages/core/src/configuration/config-init-service.ts
mahirshah: good call, updated to use yeoman generators

packages/cli/src/commands/config/init.ts
mahirshah: true, I also dont feel too strongly either way. I currently added under config:init, but feel free to move to under generate once the task/plugin generators are up.

scalvert:
@mahirshah can you rebase with the latest changes in master? There's currently a conflict. There's a few open issues here still, in case you want to fix them up!

packages/cli/__tests__/generators/generate-config-test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+describe('config-init-generator', () => {
+  it('should write a config given default answers', async () => {
+    const directory = await helpers.run(ConfigGenerator).withPrompts({
+      framework: 'Other',
+      format: CheckupConfigFormat.JSON,
+    });
+
+    assert.file('.checkuprc');
+    expect(fs.readFileSync(path.join(directory, '.checkuprc'), 'utf8')).toMatchSnapshot();
```
scalvert: I added a handy util for this stuff. Check out my generator tests.

`root.file('.checkuprc').contents` yada yada

packages/cli/__tests__/generators/generate-config-test.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('should write a config given Ember framework', async () => {
+    const directory = await helpers.run(ConfigGenerator).withPrompts({
+      framework: 'Ember',
+      format: CheckupConfigFormat.JSON,
+    });
+
+    assert.file('.checkuprc');
+    expect(fs.readFileSync(path.join(directory, '.checkuprc'), 'utf8')).toMatchSnapshot();
```
scalvert: Same as above.

packages/cli/__tests__/generators/generate-config-test.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it('should error if a checkuprc file is already present', async () => {
+    await expect(
+      helpers.run(ConfigGenerator).inTmpDir(function(dir) {
+        fs.writeFileSync(path.join(dir, '.checkuprc'), JSON.stringify({}));
```
scalvert: Same as above.

packages/core/src/generators/config/init.ts
```diff
@@ -0,0 +1,73 @@
+import * as Generator from 'yeoman-generator';
```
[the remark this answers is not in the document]
mahirshah: Hmm I think you may be looking at an old diff? The generator is now in `packages/cli/src/generators/config.ts`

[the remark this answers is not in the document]
scalvert: Yep this is an old comment.

packages/cli/__tests__/generators/generate-config-test.ts
mahirshah: updated, thanks!

PR #105, opened 2020-03-11
Converts single command to multi

scalvert:
This PR converts the single command we've been using `checkup` into a multicommand. This is in preparation for adding subcommands, such as `checkup config:init`, `checkup generate:task <taskName>`, and `checkup generate:plugin <pluginName>`.


scalvert:
This PR supersedes #103, and should be merged first.

PR #106, opened 2020-03-11
Add in card partial scaffolding

carakessler:
This adds in a partial to render a card component in the report, and modifies some existing infrastructure to allow for pdf data to be generated.

### DIFF
--- a/packages/cli/__tests__/index-test.ts
+++ b/packages/cli/__tests__/index-test.ts
@@ -6,6 +6,8 @@ import { CheckupProject, Plugin, createTmpDir, stdout } from '@checkup/test-help
 
 import cmd = require('../src');
 
+const increasedTestTimeout = 30000;
+
 describe('@checkup/cli', () => {
   describe('normal cli output with plugins', () => {
     let project: CheckupProject;
@@ -43,7 +45,9 @@ describe('@checkup/cli', () => {
                 stdout() {
                   process.stdout.write('mock task is being run\n');
                 },
-                pdf() {},
+                pdf() {
+                  return undefined;
+                },
               };
             }
           }
@@ -79,7 +83,9 @@ describe('@checkup/cli', () => {
                 stdout() {
                   process.stdout.write('mock task2 is being run\n');
                 },
-                pdf() {},
+                pdf() {
+                  return undefined;
+                },
               };
             }
           }
@@ -115,7 +121,9 @@ describe('@checkup/cli', () => {
               stdout() {
                 process.stdout.write('mock task3 is being run\n');
               },
-              pdf() {},
+              pdf() {
+                return undefined;
+              },
             };
           }
         }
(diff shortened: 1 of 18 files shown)


packages/cli/__tests__/pdf-test.ts
```diff
@@ -0,0 +1,32 @@
...
+describe('generateHTML', () => {
+  it('returns correct HTML string', async () => {
+    const htmlString = await generateHTML(mergedResults);
+
+    expect(htmlString).toMatchSnapshot();
+  });
+  it('includes tailwind UI lib and stylesheet', async () => {
+    const htmlString = await generateHTML(mergedResults);
+
+    expect(htmlString).toContain('tailwind');
```
scalvert: Should you assert more specifically on tailwind and tailwindui?

packages/core/src/types/index.ts
```diff
@@ -40,10 +40,16 @@ export type JsonTaskResult = {
...
+  stdout = 'stdout',
+  json = 'json',
+  pdf = 'pdf',
+}
+
 export interface TaskResult {
   stdout: () => void;
   json: () => JsonTaskResult;
-  pdf: () => void;
+  pdf: () => void | CardData;
```
scalvert: This is a bit of a weird intersection type. You'd typically do `CardData | undefined`, not `void`. Can we change this?

packages/cli/__tests__/pdf-test.ts
carakessler: i was waiting to start actually using tailwind to do more testing here... what did you have in mind? we arent actually doing much w tailwind yet

packages/core/src/types/index.ts
carakessler: changed

packages/cli/__tests__/pdf-test.ts
scalvert: I think specifically you're looking at whether those two CSS files are present in the report. This seems good, but I'd use more specific strings to validate this.

```
expect(html).toContain('tailwind.min.css');
expect(html).toContain('tailwind-ui.min.css');
```

carakessler: fixed :)

packages/cli/__tests__/index-test.ts
```diff
@@ -140,7 +146,7 @@ describe('@checkup/cli', () => {
       await cmd.run(['run', project.baseDir]);
 
       expect(stdout()).toMatchSnapshot();
-    }, 15000);
+    }, 30000);
```
scalvert: Should we just make this a constant so we can change it in one place?

carakessler: cute ok

packages/cli/__tests__/index-test.ts
```diff
@@ -6,6 +6,8 @@ import { CheckupProject, Plugin, createTmpDir, stdout } from '@checkup/test-help
 
 import cmd = require('../src');
 
+const increasedTestTimeout = 30000;
```
scalvert:
```suggestion
const TEST_TIMEOUT = 30000;
```

PR #108, opened 2020-03-12
Adding docs explaining scoring heuristics

carakessler:
This doc is a WIP, and I am looking for feedback here on the business logic I intend to build out for the report. There are a few different options propsed in this doc, and would love feedback on what makes the most sense to y'all!! Also if you have alternate ideas on how to "grade" our apps, please comment!!!!!!!!

[Rendered](https://github.com/checkupjs/checkup/blob/docs/docs/HEURISTICS.md)

### DIFF
--- a/docs/HEURISTICS.md
+++ b/docs/HEURISTICS.md
@@ -3,7 +3,7 @@
 ## Scoring and Heuristics (RFC - still WIP)
 
 ### Summary
-Checkups main value is in providing insights into the health of an application by helping users interpret the data we collect about an app in a way that can be understood by all of our targeted [personas](SPEC.md#Personas). 
+Checkup's main value is in providing insights into the health of an application by helping users interpret the data we collect about an app in a way that can be understood by all of our targeted [personas](SPEC.md#Personas). 
 
 For example, if a task that charts a migration is added to checkup today, and your app is 0% done with this new migration, that is totally okay. We don't expect you to be done, it only began today! The configurability of checkup is very valuable here - as the owner of a task you can provide a `failingScore` and a `passingScore` (which can be overridden via the checkup config in your application), which we use to provide a letter grade (`A`-`F`) for each task. We then do some fun math to take the grades given for each task, and the `Priority` of each task, and provide a grade for each `Category` of task, and then for the app's overall health. 
 
@@ -61,7 +61,6 @@ Alternatively, we can also just make priority a number between 1-5 or 1-10.
 ### Overall health score
 
 We also need to determine the weight of the three categories of tasks, in order to combine them into one solid health score. I propose that `Core` be weighted 2x, `Insights` 1.5x, and `Migrations` 1x. We will then take the grades of each of those sections, and multiply them out to get one final grade, or health score for the application.
-
 
 
 
--- a/docs/HEURISTICS.md
+++ b/docs/HEURISTICS.md
@@ -1,0 +1,68 @@
+# Checkup
+
+## Scoring and Heuristics (RFC - still WIP)
+
+### Summary
+Checkups main value is in providing insights into the health of an application by helping users interpret the data we collect about an app in a way that can be understood by all of our targeted [personas](SPEC.md#Personas). 
+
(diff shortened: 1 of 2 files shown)


docs/HEURISTICS.md
```diff
@@ -0,0 +1,68 @@
...
+
+For the same example above, we could also say that only `passingScore` and above count as an `A`, and only `failingScore` and below count as an `F`. So then the delta between `passingScore` and `failingScore` would be divided by 3 (`B`, `C`, `D`), making the distribution as follows: 
+
+| Score | Grade |
+|-------|-------|
+| 0-10  | A     |
+| 10-17 | B     |
+| 17-23 | C     |
+| 23-30 | D     |
+| 30-∞  | F     |
```
scalvert: Overall I feel like either of these will work, and that we should just go ahead and try one.

My feeling is that it'll become VERY apparent how these _feel_ to us when we run them. It'll likely be very obvious if we think a heuristic is reporting a value that doesn't match our expectations. 

👍

carakessler:
Based on our sync yesterday, I agree that it is not feasible to use hard counts as the passing and failing score, as these will vary greatly based on the size of the app being analyzed.

To help us come up with a better mechanism, here are some guiding questions I am using to think about this. 

1) Scenario A: I have 100 tests and 10 are skipped.
   Scenario B: I have 100,000 tests with 10,000 skipped.

Do you view A as the same as B? If A is viewed as acceptable, does that mean B is indisputably acceptable as well?   

2) If you **don't** think scenarios A && B are the same what can that mean?  
Option 1: Heuristics don't scale linearly as the app scales
Option 2: Heuristics can scale linearly up to a certain threshold, but then start to scale differently, or cap 
Option 3: Heuristics shouldn't scale at all, hard numbers are the way to go. 

3) If you think A === B here, then we just should scale linearly for disabled tests. However, are you comfortable saying that all heuristics scale the same way? I think it would be REALLY complex to have different scales, both for maintainability of the code and for easy of onboarding and writing tasks. So can you maybe think of a task that would NOT scale linearly? Looking at our initial list of tasks, will linearly work across the board? 

--------------------------------------------------------------------
For tasks without easily derivable denominators (for example, number of usages of a bad pattern, number of TODOs, number of lint violations, etc), how would we scale? 

1) Just use LOC (of a given type) as denominators, and still use percentages, they would just be really really small percentages. For example, out of the 1M lines of JS in the app, you can have 100 TODOs, so is the rule you have have .01% TODOs. 

2) If you think #1 is a good idea, you can move on. I personally think the ergonomics of defining that are poor and feel arbitrary. I want this to be "easy" for our tasks authors. 

Instead, as an author, I want to be able to sort of use my gut feeling. 

STREAM OF CONSCIOUSNESS - I am thinking here in terms of pillars (because voyager-web is my stomping ground, so its easy to conceptualize using it as basis, but the app itself is too big). So for a large pillar, like profile or feed, how many TODOs would I think are acceptable before reading their code and being like DAMN do they ever TODO anything? Okay so there are 10 engineers working on this code, and maybe each eng having 5ish TODOs assigned to them is ok, so that makes **50** a good threshold for these sorts of pillars. Yeah okay that feels good to me. So then for the medium pillars, like mynetwork or ignite, 30ish feels good. And for smaller pillars, like video or events, 10ish would feel good. 
END STREAM OF CONSCIOUSNESS

So maybe we scale apps as S/M/L (maybe use 5 levels of granularity instead aka XS, S, M, L, XL) and use hard values at each. Since we have a task that derives #of types (which I think would be more useful than LOC), we can derive what "size" your app is, (and tell you at the top - again transparency is important), and base all of your heuristics based off of that. Some apps (like v-web) may be an XXXXL, and may need to do some extra work, but thats ok (and again, we can tell people that). 

------------------------------------------------------------------------

TLDR; my proposal is that we analyze a bunch of ember apps and quantify average sizes in tshirt sizes based on type count (**not** LOC) and then scale heuristics according to size. We can have a default way of doing this (maybe doubling) and also offer additional configurability per task to define passing/failing scores per type. 

So task authoring experience would be: 

"for a medium app with _x_ components, and _y_ routes, and _z_ tests, and blah blah, what would be the passing and failing scores?"
pass: __10_____
fail: ___20____

"okay so then for a small app with _a_ components, and _b_ routes, and _c_ tests, pass would be 5 and fail would be 10. is that reasonable?" 

yes/no 

give option to define additional pass/fail numbers for small && large, if the default scale doesn't make sense.

scalvert:
@carakessler do you think we should close this in light of our current direction in terms of data schemas?

carakessler:
@scalvert i think that actions really took the place of this - closing

PR #111, opened 2020-03-13
Adding util to mock out PDF data for report

carakessler:
Adding util that returns fake data (right now just CardData because thats the only type we have) so that we can work on the report independently of the tasks. This is rough for now, will build out and formalize more as we get more types

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,1 +1,2 @@
 export { run } from '@oclif/command';
+export { generateReport } from './helpers/pdf';
--- a/packages/dev-cli/README.md
+++ b/packages/dev-cli/README.md
@@ -1,0 +1,18 @@
+# dev-cli
+
+cli to generate mock report for development
+
+[![oclif](https://img.shields.io/badge/cli-oclif-brightgreen.svg)](https://oclif.io)
+[![Version](https://img.shields.io/npm/v/plugin-default.svg)](https://npmjs.org/package/plugin-default)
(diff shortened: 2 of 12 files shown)


packages/cli/package.json
```diff
@@ -19,6 +19,7 @@
     "cli-ux": "^5.4.4",
     "date-and-time": "^0.12.0",
     "devtools-protocol": "^0.0.748504",
+    "faker": "^4.1.0",
```
mahirshah: no `@types/faker`?

packages/cli/src/utils/generate-mock-data.ts
```diff
@@ -0,0 +1,51 @@
+import { CardData, Category, Priority, TaskMetaData } from '@checkup/core';
+/*
+ To use, replace the call for `mergeTaskResults` in `outputResults()` in run.ts with `let resultsForPdf = generateMockData();`
+ You can also comment out the parts of `run()` that actually invoke the tasks to make it faster (since we aren't using the real results anyway)
+**/
```
scalvert: It's not really clear to me what we do with this stuff. Can you actually provide examples?

scalvert: Ping!

packages/cli/package.json
carakessler: added

packages/cli/src/utils/generate-mock-data.ts
carakessler: i have it working where i just comment out some code and replace existing function call with this one
would you prefer it if (as we dev) `--use-mock`  was a command line option
so we dont need to hackily comment out code?

PR #122, opened 2020-03-19
Adding generate command

scalvert:
Adds a new `generate` command, which provides infrastructure to run generators in the app.


packages/cli/__tests__/generators/__snapshots__/generate-task-test.ts.snap
```diff
@@ -0,0 +1,328 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
LLisa717: Why are we committing snapshot files?

scalvert: The snapshots are used by tests to ensure the output is what's expected. You commit these so `jest` knows that to compare against. If we didn't commit these, the tests would regenerate a new snapshot each time, making the tests meaningless.

PR #123, opened 2020-03-19
Add template lint to plugin-ember-octane

lbdm44:
Fixes: #83


packages/plugin-ember-octane/__tests__/results/octane-migration-status-tast-result-test.ts
```diff
@@ -1,14 +1,18 @@
...
-      let sampleOctaneReport = require('../__fixtures__/sample-octane-eslint-report.json');
+      let sampleESLintReport = require('../__fixtures__/sample-octane-eslint-report.json');
+      let sampleTemplateLintReport = require('../__fixtures__/sample-octane-template-lint-report.json');
       let task = new OctaneMigrationStatusTask({});
-      let taskResult = new OctaneMigrationStatusTaskResult(task.meta, sampleOctaneReport);
+      let taskResult = new OctaneMigrationStatusTaskResult(
+        task.meta,
+        sampleESLintReport,
+        sampleTemplateLintReport
+      );
```
scalvert: Can this be moved to `beforeEach`, so you can reuse these? Seems to be common setup.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
```diff
@@ -1,6 +1,9 @@
-import { BaseTaskResult, TaskMetaData, TaskResult, ui } from '@checkup/core';
-
 import { CLIEngine } from 'eslint';
+import { BaseTaskResult, TaskMetaData, TaskResult, ui } from '@checkup/core';
+import {
+  EmberTemplateLintReport,
+  EmberTemplateLintResult,
+} from '../tasks/octane-migration-status-task';
 
 // classic-decorator-hooks                  -
```
scalvert: Can we either remove this, or add it to a jsdoc comment that describes when this task result handles?

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
```diff
@@ -22,7 +25,7 @@ interface CompetionInfo {
 interface MigrationInfo {
   completionInfo: CompetionInfo;
   name: string;
-  relatedResults: CLIEngine.LintResult[];
+  relatedResults: CLIEngine.LintResult[] | EmberTemplateLintResult[];
```
scalvert: This would be a good opportunity to define an intersection type that you can use for both.

```
type LintResultCollection = CLIEngine.LintResult[] | EmberTemplateLintResult[];
```

I'm not married to the name.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
```diff
@@ -101,32 +136,72 @@ function getMigrationInfo(
   };
 }
 
+function getTemplateLintMigrationInfo(
+  migrationConfig: MigrationRuleConfig,
+  report: EmberTemplateLintReport
+): MigrationInfo {
+  // Get all files the rule applies to.
```
scalvert: Do we need this comment?

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
```diff
@@ -101,32 +136,72 @@ function getMigrationInfo(
...
+  let relatedResults = report.results.filter(({ filePath }) =>
+    migrationConfig.fileMatchers.some(fileMatcher => fileMatcher.test(filePath))
+  );
+
+  // Get all results that contain associated violations
```
scalvert: Do we need this comment?

packages/plugin-ember-octane/src/tasks/octane-migration-status-task.ts
```diff
@@ -13,8 +38,8 @@ export default class OctaneMigrationStatusTask extends BaseTask implements Task
     },
   };
 
-  public report!: CLIEngine.LintReport;
   private esLintEngine: CLIEngine;
+  private templateLinter: typeof TemplateLinter;
```
scalvert: Have you thought at all about how we can abstract these properties to make them usable by other tasks? Or do you think they're so specific to specific tasks that it doesn't make sense?

scalvert: I could imagine at the least having a simple interface that allows you to invoke the linters in a common way. What do you think?

lbdm44: Yeah, these could be easily extracted into a shared package somewhere, the reason i needed to create them for this was that ember-template-lint does not publish types.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
lbdm44: Is there a reason we need to create another type? At the moment we are not re-using it anywhere and is only needed in the MigrationInfo interface.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
lbdm44: Could you elaborate on your concern here?

At the moment, we are importing these types from where they are defined, in a follow up PR I intend to move all types to their own file to follow standard TS convention.

scalvert: Oh sorry my comment was referring to the comment listing the TODOs you had. Not sure how it got messed up.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
scalvert: Only one reason, readability. It's hard to process that type when reading it. I don't think there's anything wrong with adding types in a file for better readability.

packages/plugin-ember-octane/src/tasks/octane-migration-status-task.ts
scalvert: OK, I think we can fix that at the source. It'd be great to have these bits reusable, but it doesn't have to be done now. They can be extracted later.

packages/plugin-ember-octane/src/results/octane-migration-status-task-result.ts
lbdm44: Yeah I totally agree with splitting out types for readability, that being said I have the opposite viewpoint on this case, having another type means I need to click into it again to see that type. I'm not too strong on which was is better, but my gut jumps to not creating extra types causing more indirection.

I feel like having a pattern we always follow would be nice here, i.e. if you need have type intersections you should create a type for it and use that type wherever the intersection is needed.

PR #127, opened 2020-03-23
Adding plugin generator

scalvert:
Adds a new plugin generator.

```
checkup generate plugin my-plugin
```


scalvert:
@mahirshah I think it is?

https://github.com/checkupjs/checkup/blob/master/packages/cli/src/commands/generate.ts#L26

Is this what you mean?

scalvert:
OK, synced offline. We decided to add error handling for unrecognized generator types.

PR #140, opened 2020-03-24
Refactor types for plugin-ember-octane

lbdm44:
* Move custom types into a `types` directory.
* Remove as many custom types for Ember Template Lint as possible
* Refactor code to use ESLint's report/results API as much as possible.


packages/plugin-ember-octane/src/types/index.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export enum TemplateLintMigrationType {
+  AngleBrackets = 'angle-brackets',
+  NamedArgs = 'named-args',
+  OwnProperties = 'own-properties',
+  UseModifiers = 'use-modifiers',
+}
+
+interface CompetionInfo {
```
scalvert:
```suggestion
interface CompletionInfo {
```

packages/plugin-ember-octane/src/types/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  total: number;
+  completed: number;
+  percentage: string;
+}
+
+export interface MigrationInfo {
+  completionInfo: CompetionInfo;
```
scalvert:
```suggestion
  completionInfo: CompletionInfo;
```

packages/plugin-ember-octane/src/types/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface MigrationRuleConfig {
+  fileMatchers: RegExp[];
+  name: string;
+  rules: string[];
+}
+
+export interface EmberTemplateLintMessage {
+  rule: string;
+  severity: 0 | 1 | 2;
```
scalvert: Would this be better represented as an enum, or no?

packages/plugin-ember-octane/__tests__/results/octane-migration-status-tast-result-test.ts
```diff
@@ -2,11 +2,10 @@ import { CLIEngine } from 'eslint';
 import { stdout } from '@checkup/test-helpers';
 import { OctaneMigrationStatusTask } from '../../src/tasks';
 import { OctaneMigrationStatusTaskResult } from '../../src/results';
-import { EmberTemplateLintReport } from '../../src/tasks/octane-migration-status-task';
 
 describe('octane-migration-status-task-result', () => {
   let sampleESLintReport: CLIEngine.LintReport;
-  let sampleTemplateLintReport: EmberTemplateLintReport;
+  let sampleTemplateLintReport: CLIEngine.LintReport;
```
scalvert: Did you just make the template lint report output conform to the ESLint format?

lbdm44: Yes, yes I did, it only required transforming the output's `rule` => `ruleId`

packages/plugin-ember-octane/src/types/index.ts
lbdm44: in ESLint the severity is a type with either of those three values, I will go ahead and mirror their type defs

packages/plugin-ember-octane/__tests__/results/octane-migration-status-tast-result-test.ts
scalvert: Ya I totally understand the intent. I do feel this is a bit confusing to readers though, as we're using types from one library for another. It feels a bit like if I were borrowing your underwear...

The fact that I had to validate this with you means it's likely to be confusing to other readers.

Can we still adhere to the same structure, but not necessarily use the exact same types?

lbdm44: Discussed offline, long term we are going to publish types from Ember Template Lint, for now I am refactoring the types into their own def file so we could eventually swap them out with those published types.

PR #174, opened 2020-03-31
Changing result JSON format to decouple from category/priority ordering

scalvert:
Implementing new report JSON format as per #156 

This PR _will not include_ sorting the tasks by category/priority. A follow up PR will implement this.


packages/test-helpers/src/task.ts
```diff
@@ -0,0 +1,58 @@
...
+      },
+    };
+
+    async run() {
+      return createMockTaskResult(taskName, friendlyTaskName, category, priority, result);
+    }
+  };
+}
+
+export function createMockTaskResult(
```
carakessler: what format is this meant to be for? pdf? stdout? json? would be nice if that was clear from function name (since we will probably want `createMockPdfTaskResult` && `createMockJsonTaskResult` && createMockStdoutTaskResult`

scalvert: Ya I think we can just change the result param to be an object that represents all three results. Good callout, though.

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -7,70 +7,124 @@ tasks.someTask.isEnabled expected type boolean, but got \\"foo\\""
...
+
+=== Repository Information
+Active days:   0 days
+Age:           0 days
+Total commits: 0
+Total files:   0
+
+=== Checkup Configuration
+configHash: 47def66968059c89c59bb41f477c180c
+version:    0.0.1
```
scalvert, line 22: This piece is dynamic, so we're probably going to have to move to inline snapshots.

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -7,70 +7,124 @@ tasks.someTask.isEnabled expected type boolean, but got \\"foo\\""
...
+      \\"repository\\": {
+        \\"totalCommits\\": 0,
+        \\"totalFiles\\": 0,
+        \\"age\\": \\"0 days\\",
+        \\"activeDays\\": \\"0 days\\"
+      }
+    },
+    \\"checkup\\": {
+      \\"configHash\\": \\"47def66968059c89c59bb41f477c180c\\",
+      \\"version\\": \\"0.0.1\\"
```
scalvert, line 45: This piece is dynamic, so we're probably going to have to move to inline snapshots.

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -7,70 +7,124 @@ tasks.someTask.isEnabled expected type boolean, but got \\"foo\\""
...
+
+=== Repository Information
+Active days:   0 days
+Age:           0 days
+Total commits: 0
+Total files:   0
+
+=== Checkup Configuration
+configHash: 47def66968059c89c59bb41f477c180c
+version:    0.0.1
```
scalvert, line 106: This piece is dynamic, so we're probably going to have to move to inline snapshots.

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -7,70 +7,124 @@ tasks.someTask.isEnabled expected type boolean, but got \\"foo\\""
...
+
+=== Repository Information
+Active days:   0 days
+Age:           0 days
+Total commits: 0
+Total files:   0
+
+=== Checkup Configuration
+configHash: 052fa270c6764ec1da0fa838bbc43dde
+version:    0.0.1
```
scalvert, line 125: This piece is dynamic, so we're probably going to have to move to inline snapshots.

packages/test-helpers/src/task.ts
scalvert: I'm going to work on this in a separate PR to keep this one from getting too big.

PR #175, opened 2020-03-31
Adding classes that represent the result data for different types of tasks

carakessler:
Checkup: Task Breakdown doc was used as SoT for creating these classes.


scalvert:
This seems like an absolutely fine start to defining these. I assume we're going to iterate over these over the next few weeks, so I'm in favor of merging this.

PR #208, opened 2020-04-06
WIP: dependencies-freshness task

LLisa717:


LLisa717:
What it looks like in the console:

Screen Shot 2020-04-06 at 12 11 50 PM

scalvert:
This is a great start! Some thoughts on the output:

What if we instead determine the count of major, minor, patch dependencies that were out of date, and summarized that? 

```sh-session
=== Out-of-date Dependencies

Major: 3
Minor: 16
Patch: 13
```

Additionally, it'd be great if we could provide a `verbose` flag in the command line, which would hint to us that we can display more information. If the flag is present, we can subdivide these into 3 tables: major, minor, patch.

In order to facilitate the above, you can sort these into 3 arrays that correspond to major, minor, and patch, and to output the summary, simply use the `length` prop on the array.

Thoughts?

packages/plugin-ember/src/results/dependencies-freshness-task-result.ts
```diff
@@ -0,0 +1,67 @@
+import { BaseTaskResult, TaskResult, ui } from '@checkup/core';
+import { DepFreshnessInfo } from '../types';
+
+export default class DependenciesFreshnessTaskResult extends BaseTaskResult implements TaskResult {
+  depFreshnessInfo!: DepFreshnessInfo;
```
scalvert, line 5: I'd rename this to `outdatedDependencies`

packages/plugin-ember/src/results/dependencies-freshness-task-result.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  stdout() {
+    ui.styledHeader(this.meta.taskName);
+    ui.blankLine();
+
+    this._writeFreshnessTable(this.depFreshnessInfo);
+    ui.blankLine();
+    // Todo: should look into color schemes
```
scalvert, line 13: We're trying to keep the output simple by using oclif's UI lib. To that end, let's not focus too much on the terminal output. We want to make the reports really nice, and reserve terminal output for more infrequent usage.

packages/plugin-ember/src/results/dependencies-freshness-task-result.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  json() {
+    return {
+      meta: this.meta,
+      result: {
+        depFreshnessInfo: this.depFreshnessInfo,
```
scalvert, line 20: Again, I'd rename to `outdatedDependencies`

packages/plugin-ember/src/tasks/dependencies-freshness-task.ts
```diff
@@ -0,0 +1,37 @@
...
+  // todo: error handle when there is no outdatedDep
+  return hash({
+    tableHead: outdatedDepTable.data.head,
+    tableBody: outdatedDepTable.data.body,
+  });
+}
+
+export default class DependenciesFreshnessTask extends BaseTask implements Task {
+  meta: TaskMetaData = {
+    taskName: 'dependencies-freshness',
```
scalvert, line 23: Can we name this `outdated-dependencies`?

packages/plugin-ember/src/tasks/dependencies-freshness-task.ts
```diff
@@ -0,0 +1,37 @@
...
+    friendlyTaskName: 'Dependencies Freshness',
```
scalvert, line 24: Can we name this `Outdated Dependencies`?

packages/plugin-ember/src/tasks/dependencies-freshness-task.ts
```diff
@@ -0,0 +1,37 @@
...
+  // stripping out color legend info from the output
+  const outdatedDepTable = JSON.parse(stdout.split('}')[1] + '}}');
+  // todo: error handle when there is no outdatedDep
+  return hash({
+    tableHead: outdatedDepTable.data.head,
+    tableBody: outdatedDepTable.data.body,
+  });
+}
+
+export default class DependenciesFreshnessTask extends BaseTask implements Task {
```
scalvert, line 21: Can we name this `OutdatedDependenciesTask`?

packages/plugin-ember/src/types/index.ts
```diff
@@ -4,3 +4,8 @@ export const enum ProjectType {
   Engine = 'engine',
   Unknown = 'unknown',
 }
+
+export type DepFreshnessInfo = {
```
scalvert, line 8: If this type is only used between the task and the result, you can put it in the task and export it for use in the result.

packages/plugin-ember/src/results/dependencies-freshness-task-result.ts
```diff
@@ -0,0 +1,67 @@
+import { BaseTaskResult, TaskResult, ui } from '@checkup/core';
+import { DepFreshnessInfo } from '../types';
+
+export default class DependenciesFreshnessTaskResult extends BaseTaskResult implements TaskResult {
+  depFreshnessInfo!: DepFreshnessInfo;
+
+  stdout() {
+    ui.styledHeader(this.meta.taskName);
```
scalvert, line 8: You should use `this.meta.friendlyTaskName` here.

LLisa717:
> This is a great start! Some thoughts on the output:
> 
> What if we instead determine the count of major, minor, patch dependencies that were out of date, and summarized that?
> 
> ```
> === Outdated Dependencies
> 
> Major: 3
> Minor: 16
> Patch: 13
> ```
> 
> Additionally, it'd be great if we could provide a `verbose` flag in the command line, which would hint to us that we can display more information. If the flag is present, we can subdivide these into 3 tables: major, minor, patch.
> 
> In order to facilitate the above, you can sort these into 3 arrays that correspond to major, minor, and patch, and to output the summary, simply use the `length` prop on the array.
> 
> Thoughts?

how do you decide which one is major/minor/patch? By how many versions behind?

scalvert:
It's the delta between the `current` and `latest`. You can use the `semver` library, which has a useful `diff` function.

> diff(v1, v2): Returns difference between two versions by the release type (major, premajor, minor, preminor, patch, prepatch, or prerelease), or null if the versions are the same.

Once you get the `major`, `minor`, or `patch`, you can use that as the key to a map of the versions.

```ts
let versionTypes: Map<string, Array<string>> = new Map<string, Array<string>>([['major', []], ['minor', []], ['patch', []]]);

let versionType: string = diff(current, latest);

versionTypes.get(versionType).push(dependency); 
```

LLisa717:
other ideas: percentage of outdated dependencies, nested outdated dependencies

LLisa717:
Added outdated dependencies overview section
Screen Shot 2020-04-17 at 11 56 48 PM

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -21,6 +21,22 @@ Total files:   0
...
+Patch: 5
+
+=== Outdated Dependencies Details
+Package            Current  Wanted   Latest   Packagetype   Url             
+@types/lodash      4.14.149 4.14.150 4.14.150 checkup       devDependencies 
+@types/node        13.11.1  13.13.0  13.13.0  checkup       devDependencies 
+lint-staged        10.1.3   10.1.4   10.1.4   checkup       devDependencies 
+release-it         13.5.2   13.5.4   13.5.4   checkup       devDependencies 
+resolve            1.16.0   1.16.1   1.16.1   @checkup/core dependencies    
+yeoman-environment 2.9.1    2.9.2    2.9.2    @checkup/cli  dependencies
```
scalvert, line 38: I think the `Packagetype` and `Url` are wrong here.

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -76,5 +169,21 @@ Total files:   0
...
```
scalvert, line 186: I think the `Packagetype` and `Url` are wrong here.

packages/cli/__tests__/tasks/__snapshots__/outdated-dependencies-task-test.ts.snap
```diff
@@ -0,0 +1,113 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`outdated-dependencies-task can read task and output to console 1`] = `
+"=== Outdated Dependencies
+
+Package             Current Wanted Latest Packagetype                  Url          
+ember-template-lint 2.5.1   2.5.2  2.5.2  @checkup/plugin-ember-octane dependencies 
+eslint-plugin-ember 8.1.1   8.2.0  8.2.0  @checkup/plugin-ember-octane dependencies
```
scalvert, line 8: I think the `Packagetype` and `Url` are wrong here.

packages/cli/src/tasks/outdated-dependencies-task.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type OutdatedDependencies = {
+  tableHead: [string];
+  tableBody: Array<string[]>;
+};
+
+async function getOutdatedDep(): Promise<OutdatedDependencies> {
+  const { stdout } = await shell.exec(OUTDATED_DEP, { silent: true });
+  // stripping out color legend info from the output
+  const outdatedDepTable = JSON.parse(stdout.split('}')[1] + '}}');
```
scalvert, line 17: Can we just delete the fields from the `JSON` instead rather than trying to manipulate the JSON as a string? This feels like it could be very brittle.

packages/cli/src/tasks/outdated-dependencies-task.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const hash = require('promise.hash.helper');
+const shell = require('shelljs');
+const semver = require('semver');
+const OUTDATED_DEP = 'yarn outdated --json';
+
+export type OutdatedDependencies = {
+  tableHead: [string];
+  tableBody: Array<string[]>;
+};
```
scalvert, line 12: I think this type is fine for `stdout`, but I wouldn't output the JSON format this way. We should instead use the serialized map of `major`, `minor`, and `patch` so that it's more readable. a JSON consumer really wouldn't care about tables, which is an artifact of the way we're outputting for stdout.

packages/cli/src/tasks/outdated-dependencies-task.ts
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  async run(): Promise<TaskResult> {
+    let result: OutdatedDependenciesTaskResult = new OutdatedDependenciesTaskResult(this.meta);
+    const outdatedDep = await getOutdatedDep();
+    let versionTypes: Map<string, Array<string[]>> = new Map<string, Array<string[]>>([
+      ['major', []],
+      ['minor', []],
+      ['patch', []],
+    ]);
```
scalvert, line 42: I feel like we should destructure the array of strings into a meaningful object. I assume this is package, current, latest, wanted, packagetype, url. Can we make a strongly typed object instead? This will also be useful for the JSON output.

packages/cli/src/tasks/outdated-dependencies-task.ts
LLisa717: Actually it output is 2 JSON object, the first one is the color legend info, so had to strip out that part. I can put it in a try/catch.

PR #210, opened 2020-04-06
Adding report data infrastructure for generating HTML/PDF reports

carakessler:
Adding partials for each of the PDF component types, and setting up infrastructure to dynamically render the correct partial based on the type of data it provides


packages/cli/src/static/report-template.hbs
```diff
@@ -12,11 +12,28 @@
   <h1>Checkup report</h1>
 
   {{#each core as |priority|}}
-      {{#each priority as |taskResult|}}
```
carakessler, line 15: ill do something more elegant to get rid of some of this code duplication at some point, but works for now. apparently there is no better way to loop thru multidimensional arrays in handlebars? played around in a sandbox a bit and couldnt get anything goin either, but didnt want to spend much time here as the structure will likely change and this works for now

packages/core/src/pdf-components/pdf-component-data.ts
```diff
@@ -1,9 +1,11 @@
-import { TaskMetaData } from '../types';
+import { TaskMetaData, ReportComponentType } from '../types';
 
 export abstract class PdfComponentData {
```
scalvert, line 3: Should we name this something other than `PdfComponentData`. How about `ReportComponentData`?

scalvert: It'll also make sense with `ReportComponentType`

carakessler: sgggg

packages/cli/src/index.ts
```diff
@@ -1 +1,2 @@
 export { run } from '@oclif/command';
+export { generateReport } from './helpers/pdf';
```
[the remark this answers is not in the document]
carakessler: ah good find - that was left from something i was trying, removing

packages/core/src/report-components/report-component-data.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export abstract class ReportComponentData {
+  meta: TaskMetaData;
+  componentType: ReportComponentType;
+
+  constructor(meta: TaskMetaData, componentType: ReportComponentType) {
+    this.meta = meta;
+    this.componentType = componentType;
+  }
+}
```
scalvert, line 11: You can shorten this a bit. If you use the `public` keyword, it will automatically create the public properties and assign them.

```suggestion
export abstract class ReportComponentData {
  constructor(public meta: TaskMetaData, public componentType: ReportComponentType) {
  }
}
```

packages/core/src/report-components/table-data.ts
```diff
@@ -0,0 +1,15 @@
...
+  taskResult: DependencyResult[];
+
+  constructor(
+    meta: TaskMetaData,
+    taskResult: DependencyResult[],
+    reportComponentType: ReportComponentType = ReportComponentType.Table
+  ) {
+    super(meta, reportComponentType);
+    this.taskResult = taskResult;
+  }
```
scalvert, line 14: If you slightly reorder the params to the constructor (for consistency and clarity with the base), you can apply the `public` keyword on the last prop to remove the assignment in the constructor.

```suggestion
  constructor(
    meta: TaskMetaData,
    reportComponentType: ReportComponentType = ReportComponentType.Table,
    public taskResult: DependencyResult[]
  ) {
    super(meta, reportComponentType);
  }
```

scalvert: You can likely apply this pattern to all the classes you've defined.

packages/cli/src/static/report-template.hbs
scalvert: I think this is fine for now. It's simple and explicit.

packages/core/src/report-components/graded-table-data.ts
```diff
@@ -0,0 +1,17 @@
+import { TaskMetaData, Grade, DependencyResult, ReportComponentType } from '../types';
+import { default as TableData } from './table-data';
+
+export default class GradedTableData extends TableData {
+  grade: Grade;
+
+  constructor(meta: TaskMetaData, public taskResult: DependencyResult[]) {
```
scalvert, line 7: Should we try to unify these *-data.ts classes in terms of what they accept as constructor params?

We have the following constructor signatures:

```
// graded-table-data.ts
constructor(
	meta: TaskMetaData, 
	public taskResult: DependencyResult[]
)
// numerical-card-data.ts
constructor(
	meta: TaskMetaData, 
	public taskResult: number, 
	public resultDescription: string, 
	public resultHelp?: string
)
// pie-chart-data.ts
constructor(
    meta: TaskMetaData,
    public resultNumerator: number,
    public resultDenominator: number,
    public resultDescription: string
)
// table-data.ts
constructor(
    meta: TaskMetaData,
    reportComponentType: ReportComponentType = ReportComponentType.Table,
    public taskResult: DependencyResult[]
)
```

packages/core/src/report-components/graded-table-data.ts
```diff
@@ -0,0 +1,17 @@
...
+    super(meta, ReportComponentType.GradedTable, taskResult);
+
+    this.grade = this._deriveGrade();
+  }
+
+  _deriveGrade() {
```
scalvert, line 13: Should we just make this a lazy getter?

```
get grade() {
  return this._grade ?? (this._grade = () = > { /* do the stuff */ });
}
```

packages/core/src/report-components/graded-table-data.ts
carakessler: this is what we had in the original task breakdown doc - all the things that can be shared across all *-data.ts classes is in ReportComponentData and the rest are all unique per type of card.

carakessler: Although I think we can totally iterate and improve on this as we develop. For now, I think this is fine (especially since this is all ts and the editor tells us what we need so nicely as task authors)

packages/core/src/report-components/graded-table-data.ts
carakessler: sure that makes sense, i was planning on fleshing more of this out once this does something, since its just pretty much skeletal now, i'd like to leave it

scalvert: Sounds good. We can revisit this as you move along.

packages/core/src/report-components/graded-table-data.ts
scalvert: Ya I guess I expected us to define specific interfaces, and subsequently pass those into these classes. Or are these classes what you _return_ from the `TaskResult.pdf` calls?

carakessler: These classes would be returned directly from the task results. I don’t see how/why all these components would have the same interfaces when they don’t have the same data... if they had the same type of data they would be using the same UI component

carakessler: Correct.

scalvert: I think we're on the same page here. Going to resolve this convo.

scalvert:
@carakessler can you rebase?

PR #219, opened 2020-04-07
Adds debug package to provide useful debugging info for tasks

scalvert:
This PR adds the `debug` package, and adds the following debug info:

- `Tasks`: task instantiation and `run` start/done are also annotated
- `Config`: the config loading info is added, and the loaded config is output


packages/cli/src/task-list.ts
```diff
@@ -68,8 +71,13 @@ export default class TaskList {
...
    */
   async runTasks(): Promise<TaskResult[]> {
-    let results = await this.eachTask((task: Task) => {
-      return task.run();
+    let results = await this.eachTask(async (task: Task) => {
+      this.debug('start %s', task.constructor.name);
+
+      let result = await task.run();
+
+      this.debug('%s done', task.constructor.name);
```
lbdm44, line 79: Probably not a big deal, but i would be consistent in the debug format, i.e. `start: %s` and `done: %s`

scalvert: I'm mirroring what oclif already has in their debug info.

Example:

```
  checkup:task ProjectMetaTask created +0ms
  checkup:task CheckupMetaTask created +0ms
  checkup:task start ProjectMetaTask +0ms
  checkup:task start CheckupMetaTask +19ms
  checkup:task CheckupMetaTask done +2ms
  checkup:task ProjectMetaTask done +114ms
  @oclif/config start register-parsers hook +186ms
  @oclif/config register-parsers hook done +0ms
  @oclif/config start register-tasks hook +0ms
```

I think consistency is good here.

lbdm44: sure, I just don't agree/like how they do it in general because their logging is inconsistent

scalvert: Yep, I don't disagree. But I think in this case, consistency with what's already provided is more important than doing it a different way. There's a huge benefit to standardization, regardless of the form.

PR #221, opened 2020-04-08
Removing `undefined` from `ReportResultData`, and adding mock results to all existing tasks

carakessler:
Having `undefined` as an option for `ReportResultData` was causing problems as we needed to account for `ReportResultData` being undefined, which it never will, it was just put there as an temporary measure until we retrofitted all existing tasks to return some real data. I put in another bandaid here, ensuring that each task result returns some mock data for the pdf() hook, instead of allowing it to return undefined

### DIFF
--- a/packages/cli/__tests__/__utils__/mock-task-result.ts
+++ b/packages/cli/__tests__/__utils__/mock-task-result.ts
@@ -1,4 +1,10 @@
-import { BaseTaskResult, Category, TaskMetaData, TaskResult } from '@checkup/core';
+import {
+  BaseTaskResult,
+  Category,
+  TaskMetaData,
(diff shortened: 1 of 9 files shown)


scalvert:
The test failure is a weakness of our current plugin fixture mocking. If you change the interface/base class of tasks or results, you will likely break things due to the nature of how we're generating the fixture data dynamically for plugins. I'm going to look at fixing this now, but you'll need to ensure that the  tasks/results passed into `addTask` in `run-tests.ts` are adhering to the interface or class structure of tasks and task results.

scalvert:
I assume we'll follow this up with read implementations. Seems fine for now.

PR #222, opened 2020-04-08
Implementing numerical-card

carakessler:
Implemented numerical-card and wrote tests. Added in support to render partials for tests as well, which will be used by other partials. Screenshot below shows numerical card in the grid populated with fake data. 

Screen Shot 2020-04-07 at 4 45 00 PM

### DIFF
--- a/packages/cli/__tests__/__utils__/render-partial.ts
+++ b/packages/cli/__tests__/__utils__/render-partial.ts
@@ -1,0 +1,34 @@
+import { ReportResultData, ReportComponentType } from '@checkup/core';
+import { readFileSync } from 'fs-extra';
+import * as Handlebars from 'handlebars';
+import * as path from 'path';
+
+export type CompiledPartials = {
+  [key in ReportComponentType]: HandlebarsTemplateDelegate;
+};
+
+const compiledPartials: CompiledPartials = {
+  [ReportComponentType.NumericalCard]: getPartialDelegate('numerical-card.hbs'),
+  [ReportComponentType.Table]: getPartialDelegate('table.hbs'),
+  [ReportComponentType.GradedTable]: getPartialDelegate('graded-table.hbs'),
+  [ReportComponentType.PieChart]: getPartialDelegate('pie-chart.hbs'),
+};
+
+/*
+ * When testing partials rendering multiple times in the same module,
+ * it is efficient to compile them once in module scope and pass the precompiled
+ * partials to the `renderPartialAsHtml` function
+ **/
+function getPartialDelegate(partialPath: string): HandlebarsTemplateDelegate {
+  let fullPartialPath = path.join(__dirname, `../../src/static/components/${partialPath}`);
+  let partialRaw = readFileSync(fullPartialPath, 'utf8');
+  return Handlebars.compile(partialRaw);
+}
+
+export function renderPartialAsHtml(componentData: ReportResultData): string {
+  // TODO: remove this check once all tasks are retrofitted to return results (and undefined  is no longer a valid option for ReportResultsData)
+  if (componentData) {
+    return compiledPartials[componentData.componentType]({ taskResult: componentData });
+  }
+  return '';
+}
--- a/packages/cli/__tests__/__utils__/render-partial.ts
+++ b/packages/cli/__tests__/__utils__/render-partial.ts
@@ -14,11 +14,6 @@ const COMPILED_PARTIALS: CompiledPartials = {
   [ReportComponentType.PieChart]: getPartialDelegate('pie-chart.hbs'),
 };
 
-/*
- * When testing partials rendering multiple times in the same module,
- * it is efficient to compile them once in module scope and pass the precompiled
- * partials to the `renderPartialAsHtml` function
- **/
 function getPartialDelegate(partialPath: string): HandlebarsTemplateDelegate {
   let fullPartialPath = path.join(__dirname, `../../src/static/components/${partialPath}`);
   let partialRaw = readFileSync(fullPartialPath, 'utf8');
@@ -30,5 +25,6 @@ export function renderPartialAsHtml(componentData: ReportResultData): string {
   if (componentData) {
     return COMPILED_PARTIALS[componentData.componentType]({ taskResult: componentData });
   }
+  
   return '';
 }
--- a/packages/cli/__tests__/__utils__/render-partial.ts
+++ b/packages/cli/__tests__/__utils__/render-partial.ts
@@ -25,6 +25,6 @@ export function renderPartialAsHtml(componentData: ReportResultData): string {
   if (componentData) {
     return COMPILED_PARTIALS[componentData.componentType]({ taskResult: componentData });
   }
-  
+
   return '';
 }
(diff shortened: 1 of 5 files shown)


packages/cli/__tests__/report-components/numerical-card-test.ts
```diff
@@ -0,0 +1,25 @@
...
+  description: 'bad patterns in your app',
+};
+
+describe('the correct data is rendered for a numerical card', () => {
+  it('returns correct HTML string', async () => {
+    const htmlString = renderPartialAsHtml(
+      new NumericalCardData(expectedOutput.meta, expectedOutput.result, expectedOutput.description)
+    );
+
+    expect(htmlString).toMatchSnapshot();
```
scalvert, line 23: I was speaking to @rwjblue about this, and we both agree that moving us to using inline snapshots would be better than external snapshots. Basically, with an external snapshot you have to navigate to a separate file to validate the snapshot output, vs. the snapshot being inlined right in the test. The latter makes it easier to read, and evaluate as you're authoring/reading the tests.

You should be able to switch to `toMatchInlineSnapshot` here.

packages/cli/__tests__/__utils__/render-partial.ts
```diff
@@ -0,0 +1,28 @@
...
+ * @param jsonResult
+ * @param results
+ */
+export function renderPartialAsHtml(componentData: ReportResultData): string {
+  const partials = {
+    [ReportComponentType.NumericalCard]: 'numerical-card.hbs',
+    [ReportComponentType.Table]: 'table.hbs',
+    [ReportComponentType.GradedTable]: 'graded-table.hbs',
+    [ReportComponentType.PieChart]: 'pie-chart.hbs',
+  };
```
scalvert, line 16: Would it be better to just precompile these partials in module scope to make it faster? It doesn't really need to recompile them every time for individual tests. You can store them in a map still, but just do the precompilation as the value.

scalvert:
The screenshot looks promising, but I really can't tell what all these fake values represent. This is why I'm not a fan of faker; the data doesn't really make sense in context, and the only value that's provided is that you get to exercise the worst case scenarios for visual rendering. 

Would it be possible to update the screenshot with more realistic values?

packages/cli/__tests__/report-components/numerical-card-test.ts
carakessler: sg!

packages/cli/__tests__/__utils__/render-partial.ts
carakessler: oooh ya good idea.

packages/cli/__tests__/__utils__/render-partial.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+/*
+ * When testing partials rendering multiple times in the same module,
+ * it is efficient to compile them once in module scope and pass the precompiled
+ * partials to the `renderPartialAsHtml` function
+ **/
+export function getPartialDelegate(partialPath: string): HandlebarsTemplateDelegate {
+  const fullPartialPath = path.join(__dirname, `../../src/static/components/${partialPath}`);
+  const partialRaw = readFileSync(fullPartialPath, 'utf8');
```
scalvert, line 17: I know you've unfortunately be conditioned to this, but I'm not a fan of `const` for everything. Is there a need to make these const?

packages/cli/__tests__/report-components/numerical-card-test.ts
```diff
@@ -0,0 +1,46 @@
...
+  description: 'bad patterns in your app',
+};
+
+describe('the correct data is rendered for a numerical card', () => {
+  const compiledPartials: CompiledPartials = {
+    [ReportComponentType.NumericalCard]: getPartialDelegate('numerical-card.hbs'),
+    [ReportComponentType.Table]: getPartialDelegate('table.hbs'),
+    [ReportComponentType.GradedTable]: getPartialDelegate('graded-table.hbs'),
+    [ReportComponentType.PieChart]: getPartialDelegate('pie-chart.hbs'),
+  };
```
scalvert, line 27: Can this not just be in module scope inside the `render-partial.ts` module? I imagined we'd just build this up in module scope, and you'd just pass a key in (the type) to render.

packages/cli/__tests__/__utils__/render-partial.ts
carakessler: as opposed to `let`? why not `const`?

packages/cli/__tests__/report-components/numerical-card-test.ts
carakessler: done

packages/cli/__tests__/__utils__/render-partial.ts
scalvert: I don't want to get into the huge bikeshedding session about let vs const, but suffice it to say that I think const should be reserved for values that are true constants. These variables are incidental values in this method. 

`const` would make sense with module scoped variables, for instance, where you know the value is not intended to be mutated.

packages/cli/__tests__/__utils__/render-partial.ts
```diff
@@ -0,0 +1,34 @@
+import { ReportResultData, ReportComponentType } from '@checkup/core';
+import { readFileSync } from 'fs-extra';
+import * as Handlebars from 'handlebars';
+import * as path from 'path';
+
+export type CompiledPartials = {
+  [key in ReportComponentType]: HandlebarsTemplateDelegate;
+};
+
+const compiledPartials: CompiledPartials = {
```
scalvert, line 10:
```suggestion
const COMPILED_PARTIALS: CompiledPartials = {
```

packages/cli/__tests__/__utils__/render-partial.ts
```diff
@@ -0,0 +1,34 @@
...
+export function getPartialDelegate(partialPath: string): HandlebarsTemplateDelegate {
+  let fullPartialPath = path.join(__dirname, `../../src/static/components/${partialPath}`);
+  let partialRaw = readFileSync(fullPartialPath, 'utf8');
+  return Handlebars.compile(partialRaw);
+}
+
+export function renderPartialAsHtml(componentData: ReportResultData): string {
+  // TODO: remove this check once all tasks are retrofitted to return results (and undefined  is no longer a valid option for ReportResultsData)
+  if (componentData) {
+    return compiledPartials[componentData.componentType]({ taskResult: componentData });
```
scalvert, line 31:
```suggestion
    return COMPILED_PARTIALS[componentData.componentType]({ taskResult: componentData });
```

packages/cli/__tests__/__utils__/render-partial.ts
```diff
@@ -0,0 +1,34 @@
...
+  [ReportComponentType.GradedTable]: getPartialDelegate('graded-table.hbs'),
+  [ReportComponentType.PieChart]: getPartialDelegate('pie-chart.hbs'),
+};
+
+/*
+ * When testing partials rendering multiple times in the same module,
+ * it is efficient to compile them once in module scope and pass the precompiled
+ * partials to the `renderPartialAsHtml` function
+ **/
+export function getPartialDelegate(partialPath: string): HandlebarsTemplateDelegate {
```
scalvert, line 22: Do you need to export this? It's effectively module private, no?

scalvert, line 22: Do you need to export this? It's effectively module-level private, no?

carakessler: removed export

PR #244, opened 2020-04-13
Removing graded table template (as it will just use table.hbs) and modifying table.hbs to handled graded tables

carakessler:
After implementing graded-table.hbs, I realized that it was very simliar to table.hbs and the two could easily share a template


carakessler:
Screen Shot 2020-04-13 at 12 33 58 PM

(note - colors, and how we convey their meaning is still up in the air, sake of this PR was to dynamically render a color for background of table based on grade)

packages/cli/__tests__/report-components/graded-table-test.ts
```diff
@@ -0,0 +1,66 @@
...
+    friendlyTaskName: 'Mock Task',
+    taskClassification: {
+      category: Category.Migrations,
+      priority: Priority.Medium,
+    },
+  },
+  result: [
+    { name: 'ember-cli', value: '12.2.2', grade: Grade.A },
+    { name: 'ember-data', value: '4.2.2', grade: Grade.B },
+    { name: 'ember-qunit', value: '6.1.2', grade: Grade.D },
```
scalvert, line 23: Could we use non-Ember dependencies here, since we're in the "generic" cli package?

packages/core/src/report-components/graded-table-data.ts
```diff
@@ -1,11 +1,29 @@
...
 
+export type GradeClass = {
+  [key in Grade]: string;
+};
+const GRADE_CLASSES: GradeClass = {
+  [Grade.A]: 'bg-green-600',
+  [Grade.B]: 'bg-green-200',
+  [Grade.C]: 'bg-yellow-400',
+  [Grade.D]: 'bg-orange-500',
+  [Grade.F]: 'bg-red-600',
```
scalvert, line 12: Is there a way we can not tie classes directly to the implementation here? It feels 'disconnected'. I don't have a solution at this moment, but worth thinking about.

packages/core/src/report-components/graded-table-data.ts
```diff
@@ -1,11 +1,29 @@
...
-    super(meta, taskResult, ReportComponentType.GradedTable);
+  constructor(meta: TaskMetaData, taskResult: DependencyResult[]) {
+    taskResult.forEach(
+      (result) => (result.gradeClass = result.grade ? GRADE_CLASSES[result.grade] : '')
+    );
+    const defaultTableHeaders = Object.keys(taskResult[0]).filter(
+      (item) => item !== 'grade' && item !== 'gradeClass'
+    );
+
+    super(meta, taskResult, defaultTableHeaders);
```
scalvert, line 26: Doesn't it feel odd to mutate the taskResult in here? I assumed the taskResult class owned the data. Do we have a circular reference here? I feel like the grade should be set from the task result itself. Does it need to be set in here?

packages/core/src/report-components/graded-table-data.ts
carakessler: i agree feels odd to have tailwind css classes in this class definition. i could throw this map in a lil <script> in the partial itself (which i also considered), but i am trying to keep the partials lean and dumb, and do all the required data munging in these classes. those were the only options i could think of. 

what do you think of the tradeoff? having css implementation details in these class definitions vs keeping the templates simple and devoid of any extensive business logic?

packages/cli/__tests__/report-components/graded-table-test.ts
carakessler: will do

packages/core/src/report-components/graded-table-data.ts
carakessler: yes, the tables at this point are working only in the context of DependencyResults which is wrong; i will need to do work here so that tables aren't tied specifically to rendering deps. i have that mostly working, but wanted to keep this small and will follow up with another PR that handles the decoupling for both tables && graded tables. Which also handles this data manipulation piece. Note that grade isn't being set here, the grade class is. I wouldn't want/expect the task result to set a value related to UI of the report.

thoughts?

packages/core/src/report-components/graded-table-data.ts
carakessler: note - i am running into the same problem with the pie chart. since hbs has limited data manipulation capabilities, i am needing to convert the data i receive into the data the chartjs expects in pie-chart-data.ts

scalvert: I appreciate the desire to keep the partials clean, but I also think the classes are highly related to the partial rendering. Additionally, this keeps the styles closer to their use. I'd be in favor of merging them together.

packages/core/src/report-components/graded-table-data.ts
scalvert: Ya good point. I did notice that it was the `gradeClass`, but conflated the two. As it pertains to the display of the data, I think it's appropriate for it to be in here. As it pertains to the result itself, it should stay in the task result class.

packages/cli/__tests__/report-components/table-test.ts
```diff
@@ -45,11 +51,67 @@ describe('the correct data is rendered for a table', () => {
...
+      taskName: 'mock-task',
+      friendlyTaskName: 'Mock Task',
+      taskClassification: {
+        category: Category.Migrations,
+        priority: Priority.Medium,
+      },
+    },
+    result: [
+      { name: 'node', value: '12.2.2', grade: Grade.A },
+      { name: 'yarn', value: '4.2.2', grade: Grade.B },
```
scalvert, line 82:
```suggestion
      { name: 'foo', value: '12.2.2', grade: Grade.A },
      { name: 'bar', value: '4.2.2', grade: Grade.B },
```

I don't think we'd add `node` or `yarn` as dependencies...

packages/cli/src/static/components/table.hbs
```diff
@@ -11,9 +14,9 @@
         </tr>
       </thead>
       <tbody>
-        {{#each taskResult.taskResult as |result|}}
-          <tr class="bg-gray-100">
-            {{#each result as |row|}}
+        {{#each taskResult.resultsToRender as |resultToRender|}}
```
scalvert, line 17: Can we call this `taskData` to not confuse this with an actual `TaskResult` type? This way it's confusing, as `TaskResult` doesn't have a property called `resultsToRender`.

packages/core/src/report-components/table-data.ts
```diff
@@ -1,18 +1,52 @@
-import { TaskMetaData, ReportComponentType, DependencyResult } from '../types';
+import { TaskMetaData, Grade, ReportComponentType, DependencyResult } from '../types';
 import { ReportComponentData } from './report-component-data';
 
+export interface ResultToRender {
```
scalvert, line 4: Can we call this `ReportComponentData` or something? This name seems wrong...

packages/core/src/report-components/table-data.ts
```diff
@@ -1,18 +1,52 @@
...
+  [Grade.A]: 'bg-green-600',
+  [Grade.B]: 'bg-green-200',
+  [Grade.C]: 'bg-yellow-400',
+  [Grade.D]: 'bg-orange-500',
+  [Grade.F]: 'bg-red-600',
+};
+
 export default class TableData extends ReportComponentData {
   tableHeaders: Array<string>;
+  resultsToRender: ResultToRender[];
```
scalvert, line 22: I'd call this `reportComponentData`, or even just `data`.

packages/core/src/report-components/table-data.ts
```diff
@@ -1,18 +1,52 @@
...
   }
+
   _deriveTableHeaders(taskResult: DependencyResult[]): Array<string> {
-    return Object.keys(taskResult[0]);
+    return Object.keys(taskResult[0]).filter((item) => item !== 'grade');
+  }
+
+  _resultsContainGrades(taskResult: DependencyResult[]): Boolean {
+    return Object.keys(taskResult[0]).filter((item) => item === 'grade').length > 0;
+  }
```
scalvert, line 46: Can't we just add a getter to this class?

```
get containsGrades(): boolean {
  return this.taskResult?.pop().hasOwnProperty('grade');
}
```

scalvert: If you're going to call it repeatedly, just cache the result. 

```
get containsGrades(): boolean {
  return (this._containsGrades ?? this.taskResult?.pop().hasOwnProperty('grade'));
}
```

packages/core/src/report-components/table-data.ts
```diff
@@ -1,18 +1,52 @@
...
     this.tableHeaders = this._deriveTableHeaders(taskResult);
+
+    if (this._resultsContainGrades(taskResult)) {
+      this.grade = this._deriveGrade();
+    }
   }
+
   _deriveTableHeaders(taskResult: DependencyResult[]): Array<string> {
-    return Object.keys(taskResult[0]);
+    return Object.keys(taskResult[0]).filter((item) => item !== 'grade');
```
scalvert, line 41: You can use the getter below for this, with a `!`

packages/cli/src/static/components/table.hbs
carakessler: ya good call - its currently called `taskResult` for all components, ill change it as part of this PR for all of them

carakessler: i think i'm going to call it `resultData` though

packages/core/src/report-components/table-data.ts
carakessler: we already have a `ReportComponentData`... hmmm.... this is the the format that the result data needs to be in to be rendered in the report.. so in english it is the "result in renderable format for report" hence my `ResultToRender`. I agree its a lil awkward though, ugh naming is hard. We can just say `FormattedResult`? Is that descriptive enough?

packages/core/src/report-components/table-data.ts
carakessler: resolving, in favor of convo above (same name in question)

packages/core/src/report-components/table-data.ts
carakessler: it only called once, so no need to cache but can add getter

carakessler: also qq with the `pop().hasOwnProperty` bit.. i was thinking it may be feasible/acceptable for there to be results where some of the items have grades and some dont, which is why i was checking to see if _any_ of the items have grades, instead of just the first. so for the dependencies task, maybe we have opinions about some of the dependencies, but not all. does that feel okay to you? does that make sense?

packages/core/src/report-components/table-data.ts
carakessler: i dont think so.. one is checking if there is a grade in any of the result objects (since grades are optional), the other is returning the keys of the other key value pairs _besides_ grades, to use as headers. am i missing sometihng?

packages/core/src/report-components/table-data.ts
scalvert: Sounds good to me.

PR #246, opened 2020-04-13
Using volta to pin node && yarn versions

carakessler:


scalvert:
We should likely add @rwjblue's volta github action to ensure we're correctly setting up the right version of node/yarn within our CI jobs. We can do this later.

rwjblue:
Submitted a PR for https://github.com/checkupjs/checkup/pull/250

PR #251, opened 2020-04-14
Extracting parsers to @checkup/core package

scalvert:
Removes the `eslint` and `ember-template-lint` parsers to the core package for reuse. Also deletes the `parser-eslint` package. We will still support the dynamic loading of parsers, but these parsers are considered built-ins.


packages/core/src/parsers/ember-template-lint-parser.ts
```diff
@@ -0,0 +1,59 @@
...
+      errorCount,
+      results,
+    };
+  }
+}
+
+let createParser: CreateParser<TemplateLintConfig, Parser<Promise<TemplateLintReport>>> = function (
+  config: TemplateLintConfig
+) {
+  return new EmberTemplateLintParser(config);
```
carakessler, line 56: do we want Ember specific parsers/linters in core? or things specific to any framework here?

packages/core/src/parsers/eslint-parser.ts
```diff
@@ -0,0 +1,23 @@
...
+
+class ESLintParser implements Parser<CLIEngine.LintReport> {
+  engine: CLIEngine;
+
+  constructor(config: CLIEngine.Options) {
+    this.engine = new CLIEngine(config);
+  }
+
+  execute(paths: string[]): CLIEngine.LintReport {
+    return this.engine.executeOnFiles(paths);
```
carakessler, line 13: is this synchronous? is there an async way of doing this?

packages/core/src/parsers/registered-parsers.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { createParser as createEmberTemplateLintParser } from './ember-template-lint-parser';
+import { createParser as createEslintParser } from './eslint-parser';
+
+let registeredParsers: Map<ParserName, CreateParser<ParserOptions, Parser<ParserReport>>> = new Map<
+  ParserName,
+  CreateParser<ParserOptions, Parser<ParserReport>>
+>();
+
+registeredParsers.set('eslint', createEslintParser);
```
carakessler, line 11: are there perf concerns with registering all partials? (whether they are needed or not?)

packages/core/src/parsers/ember-template-lint-parser.ts
scalvert: Ultimately, we don't. For now, I think it's fine to move this to get the overall structure working. 

The main issue was essentially providing the parsers for tests. If we have them in their own package, we need to find a way to provide that package to the test harness in plugins that require specific parsers. I wanted to tackle this problem first, then circle back and figure that second part out.

Make sense?

packages/core/src/parsers/eslint-parser.ts
scalvert: This is sync. I can wrap it in a promise, but it seems mildly odd to do so. I thought of this myself, as consistency is quite nice. 

I'm open to doing it so that consumers can always expect an async call.

packages/core/src/parsers/registered-parsers.ts
scalvert: Partials? You've got something else on your mind :)

There's no real perf concerns as this is essentially storing a factory function that creates instances. It doesn't actually do the instantiation - that's done at the call site.

packages/core/src/parsers/eslint-parser.ts
scalvert: I've updated the eslint parser to return a promise for consistency.

packages/core/src/parsers/ember-template-lint-parser.ts
carakessler: 💯

packages/core/src/parsers/registered-parsers.ts
carakessler: LOL yes i  meant parsers. partials all up in my head. sgsg

PR #258, opened 2020-04-16
Splitting up types into more specific files

scalvert:
The single types file was getting unruly and disorganized. This change splits the types into separate files, so you can actually see what types you have.


packages/core/src/types/tasks.ts
```diff
@@ -53,6 +60,15 @@ export type JsonTaskResult = {
   result: {};
 };
 
+export type ReportResultData = NumericalCardData | TableData | GradedTableData | PieChartData;
```
carakessler, line 63: i think we will likely want to split task and results out - we dont have that many types for results yet but we also dont have any results returning data for the report yet. thoughts?

scalvert: There's a relationship between the tasks and results that I wanted to preserve here. If we strongly feel the need to split them out later we can do that, but let's just keep them together for now.

carakessler: ya we can just keep an eye on how messy it gets, and split up once there are more

PR #259, opened 2020-04-16
Adding in pie chart partial

carakessler:
Implementing view for pie charts, which includes adding in chartjs


scalvert:
This PR needs a rebase.

packages/cli/src/helpers/pdf.ts
```diff
@@ -60,7 +61,33 @@ function registerPartials() {
...
+      partial.type === ReportComponentType.PieChart
+        ? appendChartjsJsSourceFiles(readFileSync(fullPartialPath, 'utf8'))
+        : readFileSync(fullPartialPath, 'utf8');
     Handlebars.registerPartial(partial.type, partialTemplateRaw);
   });
 }
+
+function chartjsRequired(resultsForPdf: any): boolean {
+  return (
+    resultsForPdf.results.filter(
```
scalvert, line 74: Can probably use `some` here?

packages/cli/src/helpers/pdf.ts
```diff
@@ -60,7 +61,33 @@ function registerPartials() {
...
-    const partialTemplateRaw = readFileSync(fullPartialPath, 'utf8');
+    const partialTemplateRaw =
+      partial.type === ReportComponentType.PieChart
+        ? appendChartjsJsSourceFiles(readFileSync(fullPartialPath, 'utf8'))
+        : readFileSync(fullPartialPath, 'utf8');
     Handlebars.registerPartial(partial.type, partialTemplateRaw);
   });
 }
+
+function chartjsRequired(resultsForPdf: any): boolean {
```
scalvert, line 72: I'd maybe call this `requiresChart`, in case we change to a different charting lib.

packages/cli/src/helpers/pdf.ts
```diff
@@ -60,7 +61,33 @@ function registerPartials() {
...
+  return (
+    resultsForPdf.results.filter(
+      (result: ReportResultData) => result && result.componentType === ReportComponentType.PieChart
+    ).length > 0
+  );
+}
+
+function appendChartjsCssSourceFiles(reportTemplateRaw: string): string {
+  const CHART_CSS = readFileSync(path.join(__dirname, '../static/chart-bootstrap.css'), 'utf8');
```
scalvert, line 81: Bootstrap :/

And so the accumulation of the CSS libs begins...

:)

packages/core/src/types/tasks.ts
```diff
@@ -83,4 +76,22 @@ export enum ReporterType {
...
+export type ReportResultData = NumericalCardData | TableData | GradedTableData | PieChartData;
+
+export const enum ReportComponentType {
+  NumericalCard = 'numerical-card',
+  Table = 'table',
+  GradedTable = 'graded-table',
+  PieChart = 'pie-chart',
+}
+
+export interface DependencyResult {
```
scalvert, line 88: I think this type was moved from somewhere else, and may not be related as much to this PR, but the name has always struck me as too specific. It's really `GradedResult`, no?

packages/cli/src/helpers/pdf.ts
carakessler: fixed

packages/core/src/types/tasks.ts
carakessler: will fix in a future PR

PR #262, opened 2020-04-16
Refactoring build to utilize typescript project references

scalvert:
This changes restructures the project to use typescript project references that were new in 3.0. Takes advantage of package references and incremental builds.

Fixes:
- dramatically improves build times (after initial build)
- go to definition
- unblocks future PR to remove ts-jest in favor of build + test
- simplifies scripts
- fixes build watching
- moves `./src/static` to `./static` in each project

Steps after pulling these changes down:

- `git clean -fdx`
- `yarn && yarn build`
- Restart your TS server in vscode (to pick up the new references)

Subsequent builds should be much faster.


packages/cli/tsconfig.json
```diff
@@ -1,14 +1,32 @@
...
     "outDir": "lib",
     "rootDir": "src",
     "types": ["chai", "fs-extra", "jest", "node"],
     "baseUrl": ".",
     "paths": {
       "@checkup/core": ["../core/lib/index.d.ts"],
+      "@checkup/cli": ["../cli/lib/index.d.ts"],
+      "@checkup/plugin-ember": ["../plugin-ember/lib/index.d.ts"],
+      "@checkup/plugin-ember-octane": ["../plugin-ember-octane/lib/index.d.ts"],
+      "@checkup/test-helpers": ["../test-helpers/lib/index.d.ts"],
```
rwjblue, line 13: We should consider hoisting this up to the shared `tsconfig-base.json`.

packages/cli/package.json
```diff
@@ -46,7 +46,8 @@
   },
   "files": [
     "/bin",
-    "/lib"
+    "/lib",
+    "/templates"
```
rwjblue, line 50: Should add `static` here too

PR #266, opened 2020-04-17
Displaying results from meta tasks in the pdf 

carakessler:


carakessler:
Screen Shot 2020-04-17 at 4 24 27 PM

scalvert:
This is starting to come together nicely. Is the UI that crowded? We may want to change things to give it more breathing room. 

Some initial thoughts:
1. The checkup version and config hash could be smaller and using a lighter grey to reduce visual prominence (we want it as viewable info, but less prominent than the main title
1. Can we bold the package name? (travis, in this case)
1. Can we use a standard format for both of the checkup and travis version numbers? I'd propose using 

	```
	Checkup Report for travis v0.0.1
	```
	
	and
	
	```
	Checkup CLI v0.0.7
	```
1. I'd make the subtitle smaller, so that the main title has more prominence. I'd also change the font color to something more grey.

carakessler:
this whole report is not yet prettified, my plan is to get all of the data out on the paper, and then to prettify and css it up. i am thinking of it as making it "functional" as step 1, and step 2 is having it look "to spec" (when i can collaborate with people more skilled in that area). since the components arent "pretty" either, i dont see value in polishing the header at this time, because the look and feel of the header will also depend on the design patterns used in the components themselves

scalvert:
Seems like a great approach. Continue!

PR #276, opened 2020-04-20
Cleaning up types in run command

scalvert:


packages/cli/src/commands/run.ts
```diff
@@ -55,103 +70,110 @@ export default class RunCommand extends Command {
...
+  runArgs!: RunArgs;
+  runFlags!: RunFlags;
   defaultTasks: MetaTaskList = new MetaTaskList();
   metaTaskResults: MetaTaskResult[] = [];
   pluginTasks: TaskList = new TaskList();
   pluginTaskResults: TaskResult[] = [];
   checkupConfig!: CheckupConfig;
 
-  async run() {
+  public async init() {
```
carakessler, line 81: looked through a bit of oclif documentation, but am still unclear if the `init` is a lifecycle hook coming from oclif (which is wht i first assumed), or is something that needs to be invoked by a user? if it is in fact the former, why is `init` public?

scalvert: It's lifecycle. It's public so that the external command infra can invoke it, otherwise it wouldn't be invocable.

scalvert: For reference: https://github.com/oclif/command/blob/a686865b7263b7402929c56bb832908076d54452/src/main.ts#L13

carakessler: ack sg sg

PR #277, opened 2020-04-20
Splitting out pdf results into sections

carakessler:
Previously, there was a PR that went in that flattened the data structure of the results. While that works well for the JSON results, the PDF results need a bit more shape/structure so we can iterate over the different categories and priorities separately. This is useful so we can have a different section per each category and type. So basically reverted the flattening change for the pdf flow, changes the hbs files so we can render the different sections based on the new data format, and added some useful types to remove the `any` around the results coming into the pdf flow to keep things robust moving forward 

Note - none of the UI being generated right now is "to spec" - just a bunch of partials inside some section boxes with titles.

### DIFF
--- a/packages/cli/src/reporters.ts
+++ b/packages/cli/src/reporters.ts
@@ -1,17 +1,69 @@
-import { ReporterType, TaskResult, ui } from '@checkup/core';
+import {
+  ReporterType,
+  TaskResult,
+  ui,
+  Category,
+  Priority,
+  ReportComponentType,
+  UIResultData,
+  UIReportData,
+} from '@checkup/core';
 
 import { MetaTaskResult } from './types';
 import { RunFlags } from './commands/run';
 import { generateReport } from './helpers/pdf';
 
-export function _transformResults(
+export function _transformPdfResults(
   metaTaskResults: MetaTaskResult[],
-  pluginTaskResults: TaskResult[],
-  reporterType: ReporterType
+  pluginTaskResults: TaskResult[]
+): UIReportData {
+  let mergedResults: UIResultData = {
+    [Category.Insights]: {
+      [Priority.High]: [],
+      [Priority.Medium]: [],
+      [Priority.Low]: [],
+    },
+    [Category.Migrations]: {
+      [Priority.High]: [],
+      [Priority.Medium]: [],
+      [Priority.Low]: [],
+    },
+    [Category.Recommendations]: {
+      [Priority.High]: [],
+      [Priority.Medium]: [],
+      [Priority.Low]: [],
+    },
+  };
+  let requiresChart = false;
+
+  pluginTaskResults
+    .map((result) => result.pdf())
+    .forEach((taskResult) => {
+      if (taskResult) {
+        let { category, priority } = taskResult.meta.taskClassification;
+
+        mergedResults[category][priority].push(taskResult);
+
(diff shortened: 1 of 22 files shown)


packages/core/src/types/tasks.ts
```diff
@@ -62,6 +62,18 @@ export type JsonTaskResult = {
...
+  [key in Category]: {
+    [key in Priority]: ReportResultData[];
+  };
+};
+
+export type ResultsForPdf = {
+  meta: JsonMetaTaskResult;
+  results: TaskResultsForPdf;
+  requiresChart: boolean;
+};
```
scalvert, line 75: I think we should try to rename these to something more generic. We're intending to use these types to output both for PDF _and_ for HTML. If that's the case, we should maybe call the former `UIResultData` and the latter `UIReportData`. Thoughts?

packages/cli/src/reporters.ts
```diff
@@ -1,16 +1,66 @@
...
+    [Category.Migrations]: {
+      [Priority.High]: [],
+      [Priority.Medium]: [],
+      [Priority.Low]: [],
+    },
+    [Category.Recommendations]: {
+      [Priority.High]: [],
+      [Priority.Medium]: [],
+      [Priority.Low]: [],
+    },
```
scalvert, line 34: We can probably dry up this object initialization, since they're all the same shape.

packages/cli/src/reporters.ts
```diff
@@ -1,16 +1,66 @@
...
+
+  pluginTaskResults
+    .map((result) => result.pdf())
+    .forEach((taskResult) => {
+      if (taskResult) {
+        let { category, priority } = taskResult.meta.taskClassification;
+        mergedResults[category][priority].push(taskResult);
+        if (taskResult.componentType === ReportComponentType.PieChart) {
+          requiresChart = true;
+        }
```
scalvert, line 46: Nit: can we give this code more breathing room? I generally prefer to let my code breathe a bit with empty lines between statements as I find it more readable.

packages/core/src/types/tasks.ts
carakessler: good call, renaming

packages/cli/src/reporters.ts
carakessler: do you have a suggestion? i tried to dynamically create this object by specifying it needed a key from each Category and Priority, but couldnt get it to work

scalvert:
```
let mapValues = {
  [Priority.High]: [],
  [Priority.Medium]: [],
  [Priority.Low]: [],
}
```

won't work? You may need to clone...

carakessler: let categoryMapValues = {
    [Priority.High]: [],
    [Priority.Medium]: [],
    [Priority.Low]: [],
  };

  let mergedResults: UIResultData = {
    [Category.Insights]: Object.assign({}, categoryMapValues),
    [Category.Migrations]: Object.assign({}, categoryMapValues),
    [Category.Recommendations]: Object.assign({}, categoryMapValues),
  };

doesnt work because its not deep cloning, only shallow. seems like optimizing this duplication may be more code than leaving it :)

scalvert: Ya, totally not worth it. Let's just ignore for now.

PR #299, opened 2020-04-21
Adding in handlebars-helpers and making some minor changes to the template using the helpers

carakessler:


packages/cli/src/helpers/pdf.ts
```diff
@@ -9,6 +9,8 @@ import { ReportComponentType, UIReportData } from '@checkup/core';
 
 import tmp = require('tmp');
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const hbsHelpers = require('handlebars-helpers')('comparison');
```
scalvert, line 13: import not work here?

PR #313, opened 2020-04-23
Adding max height to cards and overflow scroll

carakessler:


scalvert:
Is this a stop gap plan, or something you're planning on making permanent? I ask because I'd ideally like to define what a summary looks like for cards, and ultimately where we put extended information. 

If we're potentially going to head in the direction of more fully fledged HTML as the primary source of information, we may want to reconsider how we intend to output the data. We may opt for a master detail view in lieu of cards, for instance.

carakessler:
@scalvert - stop gap plan

PR #330, opened 2020-04-28
Renaming task result methods

scalvert:
Renames the following methods in the `TaskResult` interface:

- `stdout` -> `toConsole`
- `json` -> `toJson`
- `html` -> `toResultData`


scalvert:
Just going to merge this as we already agreed to the rename offline.

PR #349, opened 2020-04-30
Adding in check to octane migration task for mixin usage

carakessler:
Mixin usage is not compatible with octane, used `no-mixin` rule from `eslint-plugin-ember` to check what percentage of files are using mixins


carakessler:
[
Screen Shot 2020-04-30 at 11 40 07 AM
](url)

scalvert:
I think we should consider something different for the title. It's more like `Mixin Removal`, right? In this case, you'd be 91% removed. What do you think?

scalvert:
Thinking more about this, if we go down the line, from top to bottom and say each title with the word **Using** in front of it, it would read like this:

- Using Native Classes
- Using Tagless Components
- Using Glimmer Components
- Using Tracked Properties
- **Using Mixin Usage**
- Using Angle Brackets Syntax
- Using Named Arguments
- Using Own Properties
- Using Modifiers

In this sense, I think we should say `Using No Mixins`. What do you think?

carakessler:
@scalvert - are you suggesting we add the word "Using" in front of each of the other checks? If not (and you are saying that is just mentally what people will be _thinking_ before each word, then you are saying instead of "Mixin Usage" it should be "No Mixins", which I feel like is questionable English but also not what this check is referring to, because its checking how many time mixins are being used, not what mixins exist. 

I think best option here is "Mixin Usage Removed"

scalvert:
> from top to bottom and say each title with the word Using in front of it

Yes, just saying it out loud as if each are being read. For consistency, I think we should try to ensure the mixins title fits with that format. 

What you've suggested above has a different format than the others - **removed** is a verb. If we think each should have a verb, so be it. I want us to be consistent in the format, from top to bottom. Right now we're not.

If we use what I said before, it would read like this:

**Using Mixin Usage Removed**

It should probably be something like

**Using No Mixins**

If you want to change to be what you suggested, we should probably change the others too.

carakessler:
Changed to "No Mixins"

PR #358, opened 2020-05-01
Refactoring tasks to include taskContext and task options"

scalvert:
This is part 1 of a 2 part PR. 

Part 1
======

- [x] Refactor tasks to take a `context` object. The `context` object will be standardized to pass in the following:

```ts
interface TaskContext {
  cliArguments: RunArgs;
  cliFlags: RunFlags;
  parsers: Map<ParserName, CreateParser<ParserOptions, Parser<ParserReport>>>;
  config: CheckupConfig;
}
```

- [x] Adding additional functionality on to the `BaseTask`
	- `enabled` getter: to determine whether the task is configured to run
	- `options` getter: to determine if any options have been passed into the config

Part 2 TODO
===========

- Get plugin name inside task
        - used to format key to lookup config options eg. ember/ember-types where "checkup-plugin-ember" is the plugin name
- Skip running tasks that are disabled
- Write tests for:
        - command with plugins (using plugin generator)
        - command with tasks with options
        - command with tasks disabled


packages/core/src/base-task.ts
```diff
@@ -1,13 +1,29 @@
...
 export default abstract class BaseTask {
-  args: any;
+  context: TaskContext;
+  meta!: TaskMetaData | TaskIdentifier;
   debug: debug.Debugger;
+  private _options!: boolean | object;
 
-  constructor(cliArguments: any) {
-    this.args = cliArguments;
+  constructor(context: TaskContext) {
```
carakessler, line 11: YES to this not being an `any`

PR #359, opened 2020-05-04
chore(test): added a pretest script to execute build prior running the tests

sarbbottam:
When I cloned the project the first thing I did after `yarn install` was `yarn test` as a reflex.
As expected tests failed with 
```
Cannot find module '@checkup/test-helpers'.
Cannot find module '@checkup/core'.
```
I later realized I should have executed `yarn build` prior to the `yarn test`.
Thus added a `pretest` in the `scripts` like so:
```
"pretest": "yarn build",
```
So that `yarn build` gets automatically executed prior to the tests, when running `yarn test`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,8 @@ When adding a new breaking change, follow this template in your pull request:
 
 1. Change into that directory, and install dependencies by running `yarn install` (this will install all dependencies for all packages via yarn workspaces):
 
+1. Test the project by running `yarn test`, it will build all the packages prior running the tests.
+
 1. Build the project by running `yarn build`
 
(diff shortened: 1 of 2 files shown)


scalvert:
Thanks for the PR! I'll have to look at this a bit, since we're using ts-jest for our tests I don't _think_ this should be an issue, but it may in fact be.

scalvert:
I'd prefer to add a build step to 'postinstall', as that's a common pattern to use. We use `ts-jest` for testing, so we don't really need a build before we test, but I'm open to bootstrapping the build.

scalvert:
This change is superseded by the `yarn prepare` script that's been added. Appreciate the PR!

PR #372, opened 2020-05-04
Part 2 of task context refactoring

scalvert:
This adds the ability to pass task options from the config through to the task, and have the task correctly pluck off the options. It also changes the overall structure of task options to better match those of eslint and others.

The new Config structure is:

```ts
type CheckupConfig = {
    plugins: string[];
    tasks: {
        [fullyQualifiedTaskName: string]: "on" | "off" | ["on" | "off", unknown];
    };
}
```

As seen above, a task config can contain either

- a `string` that denotes whether it's enabled or disabled ("on"/"off" respectively)
- a tuple of `string` and `unknown`, where the string denotes whether it's enabled or disabled ("on"/"off" respectively), and the `unknown` value is the options to be passed to the task.

The task will only parse the options it owns; it won't parse any other config values for any other tasks, thus limiting the reach tasks have into one another's configs.

## Examples

An example of disabling a configured task would be:

```json
{
  "plugins": ["checkup-plugin-ember"],
  "tasks": {
	"ember/ember-types": "off"
  }
}
```

An example of passing options to a configured task would be:

```json
{
  "plugins": ["checkup-plugin-ember"],
  "tasks": {
	"ember/ember-types": ["on", { "check-mixins": true }]
  }
}
```

TODO:

- [ ] Update config loading to normalize plugin names to `checkup-plugin-*`, so that configs can eliminate the need to specify the fully qualified plugin name.
- [ ] Pass in plugin name to task, so it can use the fully qualified plugin name to look up it's config, eg. `ember/ember-types`, where `ember` is `checkup-plugin-ember`


packages/core/__tests__/base-task-test.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.meta = {
+      taskName: 'my-fake',
+      friendlyTaskName: 'Fake',
+      taskClassification: {
+        category: Category.Insights,
+        priority: Priority.High,
+      },
+    };
+  }
+}
```
rwjblue, line 35:
```suggestion
class FakeTask extends BaseTask {
  meta = {
    taskName: 'my-fake',
    friendlyTaskName: 'Fake',
    taskClassification: {
      category: Category.Insights,
      priority: Priority.High,
    },
  };
}
```

packages/core/src/base-task.ts
```diff
@@ -2,28 +2,53 @@ import * as debug from 'debug';
...
+import { TaskConfig } from './types/configuration';
+
 export default abstract class BaseTask {
   context: TaskContext;
   meta!: TaskMetaData | TaskIdentifier;
   debug: debug.Debugger;
-  private _options!: boolean | object;
+
+  private _config!: unknown;
+  private _enabled!: string;
```
rwjblue, line 13: Can we just use private fields? Seems like TS 3.8 added support for them, and they would enforce this pretty nicely...

```suggestion
  private #config: unknown;
  private #enabled: string;
```

Some prosy bits: https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#ecmascript-private-fields

packages/core/src/base-task.ts
```diff
@@ -2,28 +2,53 @@ import * as debug from 'debug';
...
-    let options = this.context.config.tasks[this.meta.taskName];
+  private _parseConfig() {
+    if (this._config) {
+      return;
+    }
+
+    let config: TaskConfig | undefined = this.context.config.tasks[this.meta.taskName];
+
+    this._config = {};
+    this._enabled = 'on';
```
rwjblue, line 31:
```suggestion
    this.#config = {};
    this.#enabled = 'on';
```

packages/core/src/base-task.ts
```diff
@@ -2,28 +2,53 @@ import * as debug from 'debug';
...
 
-    return typeof options === 'undefined' || typeof options === 'object' || options;
+    if (typeof config === 'string') {
+      this._enabled = config;
+    } else if (Array.isArray(config)) {
+      let [enabled, taskConfig] = config;
+
+      this._enabled = enabled;
+      this._config = taskConfig;
+    }
```
rwjblue, line 40:
```suggestion
    if (typeof config === 'string') {
      this.#enabled = config;
    } else if (Array.isArray(config)) {
      let [enabled, taskConfig] = config;

      this.#enabled = enabled;
      this.#config = taskConfig;
    }
```

packages/core/src/base-task.ts
```diff
@@ -2,28 +2,53 @@ import * as debug from 'debug';
...
   }
 
-  get options() {
-    let options = this.context.config.tasks[this.meta.taskName];
+  get config() {
+    this._parseConfig();
+
+    return this._config;
```
rwjblue, line 46:
```suggestion
    return this.#config;
```

packages/core/src/base-task.ts
```diff
@@ -2,28 +2,53 @@ import * as debug from 'debug';
...
+  }
+
+  get enabled() {
+    this._parseConfig();
 
-    return typeof options === 'object' ? options : {};
+    return this._enabled === 'on';
```
rwjblue, line 52:
```suggestion
    return this.#enabled === 'on';
```

rwjblue: FWIW, it seems vaguely odd to use make `this._enabled` a string only to have `this.enabled` return a boolean

scalvert: I do agree with this, but I really don't want consumers to have to evaluate against a specific string value. Any thoughts on how this could be different?

scalvert:
I'm going to roll those TODOs over into a separate PR, to keep the scope of the PRs small.

PR #373, opened 2020-05-05
Part 3 of task context refactoring: Passing normalized plugin names to tasks

scalvert:
Continues where #372 left off.

Updates config loading to normalize plugin names to `checkup-plugin-*`, so that configs can eliminate the need to specify the fully qualified plugin name.

Passes in plugin name to task, so it can use the fully qualified plugin name to look up it's config, eg. `ember/ember-types`, where `ember` is `checkup-plugin-ember`

### DIFF
--- a/packages/checkup-plugin-ember-octane/src/hooks/register-tasks.ts
+++ b/packages/checkup-plugin-ember-octane/src/hooks/register-tasks.ts
@@ -1,8 +1,11 @@
 import { Hook } from '@oclif/config';
 import OctaneMigrationStatusTask from '../tasks/octane-migration-status-task';
+import { getPluginName } from '@checkup/core';
 
 const hook: Hook<'register-tasks'> = async function ({ context, tasks }: any) {
-  tasks.registerTask(new OctaneMigrationStatusTask(context));
+  let pluginName = getPluginName(__dirname);
+
+  tasks.registerTask(new OctaneMigrationStatusTask(pluginName, context));
 };
 
 export default hook;
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -88,8 +88,10 @@ export default class RunCommand extends Command {
   }
 
   private async registerDefaultTasks(context: TaskContext) {
-    this.defaultTasks.registerTask(new ProjectMetaTask(context));
-    this.defaultTasks.registerTask(new CheckupMetaTask(context));
+    let pluginName = 'core';
+
+    this.defaultTasks.registerTask(new ProjectMetaTask(pluginName, context));
+    this.defaultTasks.registerTask(new CheckupMetaTask(pluginName, context));
   }
 
   private async runTasks() {
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -88,7 +88,7 @@ export default class RunCommand extends Command {
   }
 
   private async registerDefaultTasks(context: TaskContext) {
-    let pluginName = 'core';
+    let pluginName = 'meta';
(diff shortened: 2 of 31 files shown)


packages/core/src/utils/plugin-name.ts
```diff
@@ -0,0 +1,31 @@
...
+  return pluginName.replace('checkup-plugin-', '');
+}
+
+export function toFullPluginName(pluginName: string) {
+  if (!PLUGIN_NAME_PATTERN.test(pluginName)) {
+    return `checkup-plugin-${pluginName}`;
+  }
+
+  return pluginName;
+}
```
rwjblue, line 17: I think these functions need to become scope aware 🤔. I fully expect `@my-org/checkup-plugin-foo` to be used. In that scenario I _think_ that `toShortPluginName` will work (resulting in `@my-org/foo`) but `toFullPluginName` won't (it would emit `checkup-plugin-@my-org/foo` as written AFAICT).

packages/checkup-plugin-ember-octane/src/hooks/register-tasks.ts
```diff
@@ -1,8 +1,11 @@
 import { Hook } from '@oclif/config';
 import OctaneMigrationStatusTask from '../tasks/octane-migration-status-task';
+import { getPluginName } from '@checkup/core';
 
 const hook: Hook<'register-tasks'> = async function ({ context, tasks }: any) {
-  tasks.registerTask(new OctaneMigrationStatusTask(context));
+  let pluginName = getPluginName(__dirname);
```
rwjblue, line 6: I don't fully grok what value `getPluginName` provides here, wouldn't the following also work?

```suggestion
  let pluginName = require('../../package').name;
```

packages/core/src/utils/plugin-name.ts
scalvert: Yep, that's true. I can extend them to be scope aware.

packages/checkup-plugin-ember-octane/src/hooks/register-tasks.ts
scalvert: It would, but I felt like this was more portable, in that we have to use it both in the hooks file, and in tests. We don't know how deep the tests are housed, so it's mildly irritating to have to think about how many directories to hop up.

packages/core/src/utils/plugin-name.ts
scalvert: I added support for scoped packages and some tests.

packages/checkup-plugin-ember-octane/src/hooks/register-tasks.ts
scalvert: Will resolve this, but feel free to reopen if you think we should discuss more.

packages/cli/src/commands/run.ts
```diff
@@ -88,8 +88,10 @@ export default class RunCommand extends Command {
   }
 
   private async registerDefaultTasks(context: TaskContext) {
-    this.defaultTasks.registerTask(new ProjectMetaTask(context));
-    this.defaultTasks.registerTask(new CheckupMetaTask(context));
+    let pluginName = 'core';
```
carakessler, line 91: why is this hardcoded here?

scalvert: Because these are internal tasks. There isn't any meaning for them to have a pluginName, as they're meta tasks. I could actually rename this to `meta`, just to make it more consistent.

scalvert: Renamed to `meta`.

PR #378, opened 2020-05-06
Correctly summarizing output of ember tasks

scalvert:
Standardizing and summarizing the Ember tasks.

Screen Shot 2020-05-05 at 5 33 00 PM

### DIFF
--- a/packages/core/src/utils/data-transformers.ts
+++ b/packages/core/src/utils/data-transformers.ts
@@ -1,4 +1,5 @@
 import * as startCase from 'startcase';
+
 import { TaskItemData } from '../types/tasks';
 /**
  * @param data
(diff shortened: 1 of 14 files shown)


packages/core/src/utils/data-transformers.ts
```diff
@@ -14,3 +16,18 @@ export function toPairs(
...
   return result;
 }
+
+export function toTaskData(results: [string, string[] | Record<string, string>][]) {
+  return results.map(([type, data]) => {
+    return toTaskItemData(type, data);
+  });
+}
+
+export function toTaskItemData(type: string, data: string[] | Record<string, string>) {
```
carakessler, line 26:
```suggestion
export function toTaskItemData(type: string, data: string[] | Record<string, string>) : TaskItemData {
```

packages/core/src/utils/data-transformers.ts
```diff
@@ -14,3 +16,18 @@ export function toPairs(
 
   return result;
 }
+
+export function toTaskData(results: [string, string[] | Record<string, string>][]) {
```
carakessler, line 20:
```suggestion 
export function toTaskData(results: [string, string[] | Record<string, string>][]) : TaskItemData[] { 
```

packages/core/src/utils/data-transformers.ts
```diff
@@ -1,3 +1,5 @@
+import * as startCase from 'startcase';
+
```
carakessler, line 2:
```suggestion
import { TaskItemData } from '../types/tasks';
```

PR #379, opened 2020-05-06
[task] TODOs

carakessler:
Adding task the checks for and counts TODOs in a repository. Since this is not living in a plugin at this time, added as a default task.


packages/cli/src/commands/run.ts
```diff
@@ -88,10 +89,9 @@ export default class RunCommand extends Command {
   }
 
   private async registerDefaultTasks(context: TaskContext) {
-    let pluginName = 'meta';
-
-    this.defaultTasks.registerTask(new ProjectMetaTask(pluginName, context));
-    this.defaultTasks.registerTask(new CheckupMetaTask(pluginName, context));
+    this.defaultTasks.registerTask(new ProjectMetaTask('meta', context));
+    this.defaultTasks.registerTask(new CheckupMetaTask('meta', context));
+    this.pluginTasks.registerTask(new TodosTask('internal', context));
```
scalvert, line 94: I wonder if we should just change this to be a standard name? Or should we try to eliminate it being needed for internal type tasks?

packages/cli/src/tasks/todos-task.ts
```diff
@@ -0,0 +1,26 @@
...
+  };
+
+  async run(): Promise<TaskResult> {
+    let result: TodosTaskResult = new TodosTaskResult(this.meta);
+    result.count = await exec(
+      `grep -r -i "TODO" ${this.context.cliArguments.path} | wc -l`,
+      {},
+      0,
+      Number
+    );
```
scalvert, line 22: I don't think we should shell out to do this search. Node has packages that let you perform string searches across file contents. I think we should use something natively node instead.

carakessler: for perf reasons? or something else? 

i saw [this](https://www.npmjs.com/package/find-in-files) and was considering using it, but after looking into it, shelling out just seemed simpler, and doesnt require adding in a new package, or doing complex file searches manually

packages/cli/src/commands/run.ts
carakessler: what is `this` here? the 'internal' plugin name? i originally was going to also create a new TaskList `internalTasks` instead of just registering under `pluginTasks`

scalvert: Sorry, "this" referred to the `pluginName`. I didn't actually realize you were registering this in the `pluginTasks` task list. I don't think we should do this.

At this point I think it's reasonable to ask whether we feel that TODO is appropriate to be an internal task, or whether there needs to be a separate place to store them. For the most part, I think this doesn't fit in with the `meta` task concept - metadata that you want to gather from the project, and display in either the header and the footer of the report. TODO doesn't fit that, so it makes sense for it to be more plugin based.

That said, your idea of having an internal tasklist that acts as internal plugins makes sense to me right now. I'd be open to the addition of this.

packages/cli/src/tasks/todos-task.ts
scalvert: The way I think about this is that if we can do this natively in node, we should. Shelling out, to me, seems like a last ditch effort if what we want to do isn't supported natively in node. I've used find-in-files before - it works just fine.

Generally, shelling out is a bit more cumbersome, and the programmatic interface is pretty limited.

carakessler: i couldn't get find-in-files working the way i wanted to, and looking at the package, it doesnt look super maintained. then i tried using `find`, but it had been archived by owner. I have it working myself by just using `fs` to read each file, and then run a regex on it on at a  time, but that seems likely less performant than grep. Do you have any recommendations for a "node" way to do it?

scalvert: I'd just do what you've described, and we can concern ourselves with perf later. Sound OK?

carakessler: yep, ill push that up shortly

packages/cli/src/commands/run.ts
carakessler: resolved offline - just adding todos-task to pluginTasks in the short term, and will soon build out an external plugin for this (and external deps task)

packages/cli/src/tasks/todos-task.ts
```diff
@@ -0,0 +1,32 @@
...
+  TaskContext,
+  Task,
+  Category,
+  Priority,
+  TaskMetaData,
+  FileSearcherTask,
+} from '@checkup/core';
+import TodosTaskResult from '../results/todos-task-result';
+
+export default class TodosTask extends FileSearcherTask implements Task {
```
scalvert, line 12: I feel like we should start thinking about naming tasks more what they do, like what eslint does. In that sense, we should think about naming this `FindTodosInCommentsTask` and the file would be `find-todos-in-comments-task.ts`. This makes it much easier to understand what the task does just by reading the filename. We should also go through each current task and make sure its name makes sense.

Thoughts?

packages/core/__tests__/searchers/file-searcher-test.ts
```diff
@@ -0,0 +1,90 @@
...
+describe('file-searcher', () => {
+  let project: Project;
+
+  beforeEach(function () {
+    project = new Project('foo', '3.1.4');
+    project.files = {
+      node_modules: {
+        'lion.js': `
+          function growl() {
+            // TEST
```
scalvert, line 14: Are these meant to be `TODO`?

packages/core/src/searchers/file-searcher.ts
```diff
@@ -32,26 +36,67 @@ export default class FileSearcher {
...
+          .then((fileString: string) => {
+            for (let searchPatternName in this.searchPatterns) {
+              countData[searchPatternName] += this._getStringSearchCount(
+                fileString,
+                this.searchPatterns[searchPatternName]
+              );
+            }
+          })
+          .catch(() => {
+            debug(`Cant read file ${file}`);
```
scalvert, line 88: Do we actually want to fail/notify here instead? Typically we use `debug` to output useful information when using that env var, such as a list of all the files we're searching. We wouldn't normally use it in a catch block.

packages/core/__tests__/searchers/file-searcher-test.ts
carakessler: nah - this is a test for the file searcher itself, not the todo task.

packages/core/src/searchers/file-searcher.ts
carakessler: hmm so generally what was being caught here was an attempt to read a dir, since walksync returns dirs, as well as plain files. so it would have `['foo/bar', 'foo/bar/blah.js']`. Should I just swallow these?

packages/cli/src/tasks/todos-task.ts
carakessler: makes sense to me! i agree that sounds more descriptive. i can just push this task name as is, then push up a follow up RB that renames all tasks

scalvert: Sounds good. Do you want to create an issue to ensure we follow up?

packages/core/__tests__/searchers/file-searcher-test.ts
scalvert: 👍

packages/core/src/searchers/file-searcher.ts
scalvert: So what would you expect to happen if you were to execute this? Ignoring errors would lead to inconsistent results, so I think we should throw and ultimately error in Checkup. 

Perhaps this is a good example of Task error isolation - we can fail a single task, but shouldn't fail the whole checkup run. What do you think?

packages/cli/src/tasks/todos-task.ts
carakessler: https://github.com/checkupjs/checkup/issues/406

packages/core/src/searchers/file-searcher.ts
carakessler: so i would say for tasks that are doing a string search on the app shouldn't necessarily fail if a single file cant be read.. i can filter the results from walksync so the _known_ catches will no longer happen, but regardless, string searching in this manner is pretty brittle, and if there is a file that cant be read in the repo, i dont know that we would want to throw an error. maybe we can add a count/list of files that count be parsed, but not actually fail the task?

scalvert:
I think `FileSearcher.ts` needs some cleanup and consistency. I think this class can actually turn into a simple module with two functions that are exported.

I propose:

```
export async function findInFiles(paths?: string[]): Promise<FileSearchResults>
```

and

```
export async function findFiles(patterns: Record<string, string[]>): Promise<FileSearchResults>
```

```
interface FileSearchResults {
  results: Record<string, string[]>;
  errors: string[];
}
```

Now, I _know_ the return types won't match with our current types and that will take a small amount of work to clean up, but I think we should try to make the return type as flexible as possible (potentially using union types or generics to make them flexible) so that we have the cleanest, simplest signatures for these functions.

carakessler:
`findInFiles` would also need to take in `patterns: Record<string, string[]>` as a parameter. What are you proposing the results datatype consists of for `findInFiles`? (we did just discuss offline that leaving the count for files is best for now, and if we see people need more indepth info, we could iterate). 

Also - are you proposing we do away with `FileSearcherTask` as part of this? If so, I definitely think this should happen as a followup PR

scalvert:
> Also - are you proposing we do away with FileSearcherTask as part of this?
yes

> What are you proposing the results datatype consists of for findInFiles?
That's an exercise for the implementer; to figure out how to cleanup the two paired function calls so that their signatures are more consistent. It will make it much easier to use them if they're called in a similar fashion, and return a similar shaped result (object with results and errors props)

carakessler:
@scalvert - are you proposing that refactor be a part of this PR? seems outside the scope (as this is already sort of large).

scalvert:
Can you just pull out the changes to file-searcher.ts, make those, and push a separate PR? That will let you leave this one as is for now, and you can rebase once those changes are merged.

carakessler:
Discussed offline - changes to file-searcher will be made separately and tracked here https://github.com/checkupjs/checkup/issues/423

PR #386, opened 2020-05-06
[task] - Ember test types

carakessler:
Adding task the breaks down the types of tests in an ember application


carakessler:
Screen Shot 2020-05-06 at 11 12 26 AM
Screen Shot 2020-05-06 at 11 12 40 AM

scalvert:
Neat :)

scalvert:
Should we just build this one out more and add `skip` too? That way we can show the total, and the percent disabled?

carakessler:
oooh cute idea @scalvert - and we can have percentage disabled broken down per test type? i still think we can go ahead and merge this (unless you have concerns) and then iterate from there, but i think it makes sense for those pieces of data to be grouped together, yes

scalvert:
I'm down with merging and expanding the scope of this task later!

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -0,0 +1,47 @@
...
+  testTypes!: TaskItemData[];
+
+  findByType(typeName: string): TaskItemData | undefined {
+    return this.testTypes.find((type) => type.type === typeName);
+  }
+
+  toConsole() {
+    ui.styledHeader(this.meta.friendlyTaskName);
+    ui.blankLine();
+    ui.table(this._deriveTestPercentage(this.testTypes), { type: {}, total: {}, percentage: {} });
```
scalvert, line 13: I've been using the object argument for the headers to provide a capitalized header string for more consistent output. Can we do that?

```suggestion
    ui.table(this._deriveTestPercentage(this.testTypes), { type: { header: 'Test Type' }, total: { header: 'Total' }, percentage: { header: '' } });
```

I left the last blank as I'm not sure we need a header in that column?

packages/checkup-plugin-ember/src/tasks/ember-test-types-task.ts
```diff
@@ -0,0 +1,32 @@
+import { Category, FileSearcherTask, Priority, Task, TaskContext, TaskResult } from '@checkup/core';
+
+import EmberTestTypesTaskResult from '../results/ember-test-types-task-result';
+
+// NOTE: this task gets number of test files, not individual test functions
```
scalvert, line 5: I wonder if we should be counting the unit, rendering, and application test patterns instead? This isn't giving an accurate count of the tests, only the modules, right? We could add some lint rules.

carakessler: Yeah, I thought this may be confusing, which is why I called out that this isnt counting the actual number of tests... I think that number of modules as the main "unit" here to count makes the most sense, but I agree that number of individual tests would also be useful. I think that I can rename this to specifically say  # of test modules, and then use linting to get the number of tests && skips to expand this task. thoughts?

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
carakessler: the ui table already capitalizes the headers... if you look at the screenshot of the output I have above, they are already capitalized

packages/checkup-plugin-ember/src/tasks/ember-test-types-task.ts
scalvert: Ya I actually think the module count isn't useful at all. Why would it be?

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -0,0 +1,47 @@
...
+    );
+    return testTypes.map((testType) => {
+      testType.percentage = this._formatAsPercetange(testType.total / totalTests);
+      return testType;
+    });
+  }
+
+  _formatAsPercetange(value: number) {
+    return `${(value * 100).toFixed()}%`;
+  }
```
scalvert, line 46: I could see this being useful in other tasks. WDYT?

packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
```diff
@@ -0,0 +1,127 @@
...
+    expect(stdout()).toMatchSnapshot();
+  });
+
+  it('returns all the test types (including nested) found in the app and outputs to the console', async () => {
+    project.files = Object.assign(project.files, {
+      'index.js': 'index js file',
+      addon: TESTS,
+    });
+    project.addInRepoAddon('ember-super-button', 'latest');
+    // @ts-ignore
```
scalvert, line 71: Why do we need `@ts-ignore`s here and below? If the type isn't being derived correctly, we should fix this.

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
scalvert, line 13: We should update this to use `ui.section`

scalvert: Oh so it does. I wonder why it wasn't capitalizing on my task. Will have to check.

scalvert:
> I'm down with merging and expanding the scope of this task later!

I actually retract this statement :)

I don't think doing a count of modules is useful for us at all (unless you can think of a reason it would be). We need to get a count of tests of the different types, and then additionally count `skip` and `only`. We'd then end up with:

- `test`, `skip` and `only` (unit)
- `test`, `skip` and `only` (rendering by detecting `setupRenderingTest`)
- `test`, `skip` and `only` (application by detection `setupApplicationTest`)

I think we can build a purpose-built lint rule to accommodate this.

scalvert:
Here's a sample rule we could use for this (needs tests):

```js
export default function(context) {
  let testType = 'unit';
  return {
    'CallExpression > Identifier[name=setupRenderingTest]'(node) {
      testType = 'rendering';
    },
    'CallExpression > Identifier[name=setupApplicationTest]'(node) {
      testType = 'application';
    },
    'CallExpression > Identifier[name=test]'(node) {
      context.report({
        node,
        message: `Test type ${testType} found`
      });
    },
    'CallExpression > Identifier[name=skip]'(node) {
      context.report({
        node,
        message: `Test type ${testType} found with skip`
      });
    }
  };
};
```

gabrielcsapo:
`context.report` can also support data being passed in as well https://eslint.org/docs/developer-guide/working-with-rules#context-report.

packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
carakessler: I copy pasted that from what you did in this test (packages/checkup-plugin-ember/__tests__/ember-types-task-test.ts) - it seems like ts thinks that the addon can be undefined without that there

carakessler:
Screen Shot 2020-05-18 at 5 47 16 PM

packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
scalvert: I think we can do 

```ts
project.files.lib['ember-super-button']!.addon = TESTS;
```

no?

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -0,0 +1,40 @@
...
+  testTypes!: TestTypeInfo[];
+
+  toConsole() {
+    ui.section(this.meta.friendlyTaskName, () => {
+      ui.table(this.testTypes, {
+        type: {},
+        tests: {},
+        skipped: {},
+        percentageSkipped: { header: '' },
+      });
```
scalvert, line 14: Do you think a table is the best output format for this? What if we did a similar output breakdown like I did for dependencies?

You could even use a bar format for each test type:

```
Unit
■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 329
■ test (300) ■ skip (25) ■ only (4)
```

Obviously I can't represent the different colors on here, but you get the idea.

packages/checkup-plugin-ember/src/lint-rules/test-type-counts/test-type-counts.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
scalvert, line 1: Could we put this in a directory called `eslint/rules`. We can also colocate the tests in that same folder so they don't oddly intermingle with the other jest tests. Thoughts? 

I would also name this `test-types.js`. The counting is actually happening outside of this rule.

packages/checkup-plugin-ember/__tests__/test-type-counts-test.js
```diff
@@ -0,0 +1,56 @@
+const rule = require('../src/lint-rules/test-type-counts/test-type-counts'),
```
scalvert, line 1: Can we colocate this test under an `eslint/tests` directory? This way we encapsulate the eslint stuff in one place rather than scattering it around the other code.

scalvert:
@gabrielcsapo the `data` property passed to `context.report` is for placeholder data that's intended to be used by the message property. We can likely use this to output the type to the message itself, but I'm not sure it's propagated to the  eslint report. If it was, it'd make it much easier to parse this content!

Will investigate.

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
carakessler: what is "only" representing in that bar graph (or is that just a copy/pasta error)? 

Hmmm okay thinking about it. I think that the table is the best format for the data that I posted screenshot of above (aka # of tests enabled vs disabled), but I think it may be cool to also have a bar with the breakdown of the different types of test to sort of show testing pyramid (but in a bar). So there would still be the table I posted, but in _addition_ maybe a bar with: 
```
Total tests
■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 329
■ unit (100) ■ rendering (55) ■ application (174)
```

Thoughts?

carakessler:
@scalvert - yes, @gabrielcsapo and I looked into this offline. It seems like there is no way to send structured data as part of an eslint result. Would make things much more fun!!!! They strip out the `data` field, and just use it for string interpolation (seems quite useless imo)

scalvert:
@carakessler and @gabrielcsapo thanks for looking into this. I suspected as much!

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
scalvert: `only` is [an additional API](https://api.qunitjs.com/QUnit/only) in QUnit. We likely want to prevent the use of this, as it means that it's skipping the other tests in the module.

The reason I suggested the above output is because we want the full picture of the test breakdown, by type, and secondarily by API (`test`, `skip`, `only`). If we go with your approach it'll be lossy, in that we won't be able to see the API breakdown.

gabrielcsapo:
Yeah what's weird is they use the data to assert against in tests to make sure placeholders and the data behind is left over. It should be possible to open an issue against eslint to use this in reporters like this one.

scalvert:
@gabrielcsapo I think in this case we're abusing the eslint API to do something it's not intended for. I'd have to think a bit more about whether it makes sense for this functionality to be integrated into eslint.

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
carakessler: added `only` and `todo` (but they are only rendered to console if their values are >0 for some type of test). new output looks like this (assuming there are onlys and todos for the app being analyzed, otherwise those columns are omitted): 

```
=== Ember Test Types ===
Type           Tests   Only    Todo    Skipped
unit           681     0       0       12      2%
rendering      1062    0       0       7       1%
application    956     1       1       67      7%
=== Test Type Breakdown
■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 2699 tests
■ unit (681)  ■ rendering (1062)  ■ application (956)


```

gabrielcsapo:
I would agree, I would acorn for this type of pattern finding as it is better suited for being able to pass around structured data.

packages/checkup-plugin-ember/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig-base.json",
   "compilerOptions": {
     "outDir": "lib",
+    "allowJs": true,
```
scalvert, line 5: Could we instead convert the js files to TS for the eslint stuff?

packages/core/src/utils/ui.ts
```diff
@@ -53,7 +53,7 @@ export const ui = Object.assign(ux, {
     const completedSegments = Math.ceil(complete / barSegment);
     const incompleteSegments = maximum - completedSegments;
     const bar = `${chalk.green(barTick.repeat(completedSegments))}${chalk.grey(
-      barTick.repeat(incompleteSegments)
+      barTick.repeat(incompleteSegments < 0 ? 0 : incompleteSegments)
```
scalvert, line 56: You could also do `Math.max(0, maximum - completedSegments)` on line 54 to default this to 0 if negative.

packages/checkup-plugin-ember/__tests__/__snapshots__/ember-test-type-task-test.ts.snap
```diff
@@ -0,0 +1,121 @@
...
+
+Type           Tests   Only    Todo    
+unit           0       0       0       
+rendering      1       1       1       
+application    0       0       0       
+
+=== Test Type Breakdown
+
+■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 1 tests
+■ unit (0)  ■ rendering (1)  ■ application (0)
```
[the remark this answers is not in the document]
carakessler: ya i was back and forth on this - wasnt sure what was the most intuitive. i can make the bar graph totals

packages/checkup-plugin-ember/tsconfig.json
carakessler: discussed offline - leaving this in for now, and isolating js files to `eslint` dir

packages/core/src/utils/ui.ts
carakessler: ... is that better? the ternary is more performant (not that performance really matters here but still) and does the same thing...

packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
carakessler: fixed

carakessler: that didnt work, but `(project.files.lib as any)['ember-super-button'].addon = TYPES;` did

packages/core/src/utils/ui.ts
scalvert: I'm not really thinking about performance. It just reads better so that lines 53 and 54 seem cohesive when we're calculating.

carakessler: i kept the `incompleteSegments` var there for readability and made the update. don't necessarily think this is an improvement, but not a hill i am willing to die on

packages/checkup-plugin-ember/src/utils/transformers.ts
```diff
@@ -0,0 +1,99 @@
...
+        formattedCounts.skipUnit +
+        formattedCounts.testUnit,
+      onlys: formattedCounts.onlyUnit,
+      todos: formattedCounts.todoUnit,
+      skips: formattedCounts.skipUnit,
+      tests: formattedCounts.testUnit,
+      percentageSkipped: derivePercentageString(
+        formattedCounts.skipUnit,
+        formattedCounts.testUnit + formattedCounts.skipUnit
+      ),
```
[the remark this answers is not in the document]
scalvert: FYI I'm working up an alternative. Should have something pushed up soon.

PR #387, opened 2020-05-07
[task] - number of Ember engines and addons

carakessler:
Returns # engines and # addons

also fixed bug in generator fixture - missing parenthesis

### DIFF
--- a/packages/checkup-plugin-ember/src/results/ember-in-repo-addons-engines-task-result.ts
+++ b/packages/checkup-plugin-ember/src/results/ember-in-repo-addons-engines-task-result.ts
@@ -1,0 +1,34 @@
+import { BaseTaskResult, TaskResult, ui, TableData } from '@checkup/core';
+
+export default class EmberInRepoAddonEnginesTaskResult extends BaseTaskResult
+  implements TaskResult {
+  inRepoAddons!: string[];
+  inRepoEngines!: string[];
+
+  toConsole() {
+    if (!this.inRepoAddons && !this.inRepoEngines) {
+      return;
+    }
+
+    ui.section(this.meta.friendlyTaskName, () => {
+      ui.log(`In-Repo Addons: ${this.inRepoAddons.length}`);
+      ui.log(`In-Repo Engines: ${this.inRepoEngines.length}`);
+    });
+  }
+
+  toJson() {
+    return {
+      meta: this.meta,
+      result: { inRepoAddons: this.inRepoAddons, inRepoEngines: this.inRepoEngines },
+    };
+  }
+
+  toReportData() {
+    return [
+      new TableData(this.meta, [
+        { name: 'In-Repo Addons', value: this.inRepoAddons.length },
+        { name: 'In-Repo Engines', value: this.inRepoEngines.length },
+      ]),
+    ];
+  }
+}
--- a/packages/test-helpers/src/ember-cli-fixturify-project.ts
+++ b/packages/test-helpers/src/ember-cli-fixturify-project.ts
@@ -2,6 +2,11 @@
 
 import CheckupFixturifyProject from './checkup-fixturify-project';
 
+const enum InRepoPackageType {
+  Addon = 'addon',
+  Engine = 'engine',
+}
+
 const Project = require('fixturify-project');
 
 /**
@@ -35,12 +40,22 @@ export default class EmberCLIFixturifyProject extends CheckupFixturifyProject {
   }
 
   addInRepoAddon(name: string, version = '0.0.0') {
-    const inRepoAddon = new Project(name, version, (project: any) => {
+    this.createPackage(name, version, InRepoPackageType.Addon);
+  }
+
+  addInRepoEngine(name: string, version = '0.0.0') {
+    this.createPackage(name, version, InRepoPackageType.Engine);
+  }
+
+  private createPackage(name: string, version = '0.0.0', type: InRepoPackageType) {
+    const inRepoPackage = new Project(name, version, (project: any) => {
       project.pkg.keywords.push('ember-addon');
+      if (type === InRepoPackageType.Engine) {
+        project.pkg.keywords.push('ember-engine');
+      }
       project.pkg['ember-addon'] = {};
       project.files['index.js'] = 'module.exports = { name: require("./package").name };';
     });
-
     // configure the current project to have an ember-addon configured at the appropriate path
     let addon: any = (this.pkg['ember-addon'] = this.pkg['ember-addon'] || {});
     addon.paths = addon.paths || [];
@@ -55,6 +70,6 @@ export default class EmberCLIFixturifyProject extends CheckupFixturifyProject {
     this.files.lib = this.files.lib || {};
 
     // insert inRepoAddon into files
-    Object.assign(this.files.lib, inRepoAddon.toJSON());
+    Object.assign(this.files.lib, inRepoPackage.toJSON());
   }
 }
--- a/packages/checkup-plugin-ember/__tests__/__snapshots__/ember-in-repo-addons-engines-task-test.ts.snap
+++ b/packages/checkup-plugin-ember/__tests__/__snapshots__/ember-in-repo-addons-engines-task-test.ts.snap
@@ -1,0 +1,69 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ember-in-repo-addons-engines-task can read task and output to console 1`] = `
+"=== Ember In-Repo Addons / Engines ===
+
+In-Repo Addons: 2
+In-Repo Engines: 2
+
+"
+`;
+
+exports[`ember-in-repo-addons-engines-task can read task as JSON 1`] = `
+Object {
+  "meta": Object {
+    "friendlyTaskName": "Ember In-Repo Addons / Engines",
+    "taskClassification": Object {
+      "category": "insights",
+      "priority": "low",
+    },
+    "taskName": "ember-in-repo-addons-engines",
+  },
+  "result": Object {
+    "inRepoAddons": Array [
+      "admin",
+      "shopping-cart",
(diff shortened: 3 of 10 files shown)


packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
```diff
@@ -0,0 +1,50 @@
...
+  TaskResult,
+  FileSearcherTask,
+  TaskContext,
+} from '@checkup/core';
+import EmberAddonEngineCountTaskResult from '../results/ember-addon-engine-count-task-result';
+import { PackageJson } from 'type-fest';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+
+export default class EmberAddonEngineCountTask extends FileSearcherTask implements Task {
```
scalvert, line 15: Can we rename this `EmberInRepoAddonEnginesTask`? I think it's important to distinguish that we're actually looking at in-repo, not addons/engines from external sources. I also don't want to include count as part of the name in case we want to add to the data we're gathering, such as adding dependency checks for the in-repo packages.

packages/checkup-plugin-ember/__tests__/ember-addon-engine-count-task-test.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('ember-addon-engine-count-task', () => {
+  let emberProject: EmberProject;
+  let pluginName = getPluginName(__dirname);
+
+  beforeEach(() => {
+    emberProject = new EmberProject('checkup-app', '0.0.0', (emberProject) => {
+      emberProject.addDependency('ember-cli', '^3.15.0');
+    });
+    emberProject.addInRepoAddon('ember-cli-blueprint-test-helpers', 'latest');
+    emberProject.addInRepoAddon('ember-cli-shmooprint-test-helpers', 'latest');
```
scalvert, line 15: I don't think we'd have in-repo addons named like this. Can we just add something that seems more realistic?

```suggestion
    emberProject.addInRepoAddon('admin', 'latest');
    emberProject.addInRepoAddon('shopping-cart', 'latest');
```

packages/checkup-plugin-ember/src/results/ember-addon-engine-count-task-result.ts
```diff
@@ -0,0 +1,33 @@
+import { BaseTaskResult, TaskResult, ui, TableData } from '@checkup/core';
+
+export default class EmberAddonEngineCountTaskResult extends BaseTaskResult implements TaskResult {
+  addons!: number;
+  engines!: number;
```
scalvert, line 5: Should we be gathering more data than just counts? It may be usedful to have the package names, etc.

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    result.addons = 0;
+    result.engines = 0;
+
+    let packageJsonPaths: string[] = searchResults[0].data as string[];
+
+    packageJsonPaths.forEach((pathName: string) => {
+      let packageJson: PackageJson = fs.readJsonSync(
+        path.join(path.resolve(this.context.cliArguments.path), pathName)
+      );
```
scalvert, line 41: We have a `getPackageJson` util in core. Will that work?

packages/test-helpers/src/ember-cli-fixturify-project.ts
```diff
@@ -2,6 +2,11 @@
 
 import CheckupFixturifyProject from './checkup-fixturify-project';
 
+const enum PackageType {
```
scalvert, line 5:
```suggestion
const enum InRepoPackageType {
```
?

packages/test-helpers/src/ember-cli-fixturify-project.ts
```diff
@@ -35,12 +40,22 @@ export default class EmberCLIFixturifyProject extends CheckupFixturifyProject {
...
-    const inRepoAddon = new Project(name, version, (project: any) => {
+    this.createPackage(name, version, PackageType.Addon);
+  }
+
+  addEngine(name: string, version = '0.0.0') {
+    this.createPackage(name, version, PackageType.Engine);
+  }
+
+  private createPackage(name: string, version = '0.0.0', type: PackageType) {
+    const newPackage = new Project(name, version, (project: any) => {
```
scalvert, line 51:
```suggestion
    const inRepoPackage = new Project(name, version, (project: any) => {
```

packages/test-helpers/src/ember-cli-fixturify-project.ts
```diff
@@ -55,6 +70,6 @@ export default class EmberCLIFixturifyProject extends CheckupFixturifyProject {
     this.files.lib = this.files.lib || {};
 
     // insert inRepoAddon into files
-    Object.assign(this.files.lib, inRepoAddon.toJSON());
+    Object.assign(this.files.lib, newPackage.toJSON());
```
scalvert, line 73:
```suggestion
    Object.assign(this.files.lib, inRepoPackage.toJSON());
```

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
carakessler: aye aye captain

packages/checkup-plugin-ember/src/results/ember-addon-engine-count-task-result.ts
carakessler: I personally don't think listing that out would be particularly useful as a checkup insight. What do you think would be the use case for that info?

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
carakessler: hmmm not _really_... what I am doing here is first finding all `package.json`s in the app, and then reading them. That util just takes in  a root, finds the package.json and reads it. I suppose that util could be expanded to do what I doing here too, but that doesn't seem reusable, plus its pretty trivial

packages/checkup-plugin-ember/src/results/ember-addon-engine-count-task-result.ts
scalvert: There's a certain large app I'm thinking of that has numerous engines/addons. It'd be great to be able to gauge the distribution of ownership of them, and the only way to do that is to have the package names. I'm not suggesting outputting that for the summary, but gathering that data for verbose output. Thoughts?

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
scalvert: Ya I know it's not doing _precisely_ what we want, but it _does_ take a path, and it would be useful to have a util that may be more flexible for future Task authors. Considering we have a number of tasks that read package.json files, this seems like a good change to make.

carakessler: Are you suggesting that I have that util leverage the file searcher, and search the repo for all package.jsons? Or have the util take in an array of package.jsons, and just have it read them? 

I guess if you are suggesting the former, I suppose that could make sense as a util, it just feels pretty different than what the current one is doing. I could maybe write a separate function in that util module that does that, but again, it isnt really doing much, just using FileSearcher and then using `fs` to read it. If you are suggesting the latter, I think that feels like just a wrapper on `fs.readJsonSync`, which feels like an unnecessary abstraction. 

Thoughts?

packages/checkup-plugin-ember/src/results/ember-addon-engine-count-task-result.ts
carakessler: Ya agreed. I am going to make that change now, I will collect the names and then just [].length them in the results for stdout and ui report, but I will throw them in the JSON.

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
scalvert: This particular snippet reads a single package.json file. All I'm suggesting is we utilize something that's already purpose built for that, and adjust it as necessary to support additional use cases.

packages/checkup-plugin-ember/src/results/ember-addon-engine-count-task-result.ts
```diff
@@ -0,0 +1,34 @@
+import { BaseTaskResult, TaskResult, ui, TableData } from '@checkup/core';
+
+export default class EmberInRepoAddonEnginesTaskResult extends BaseTaskResult
```
scalvert, line 3: You renamed the class but not the file. Can we rename all the files too?

packages/checkup-plugin-ember/src/tasks/ember-addon-engine-count-task.ts
carakessler: okay, i can abstract out that piece from existing util and leverage it in both

packages/checkup-plugin-ember/src/results/ember-in-repo-addons-engines-task-result.ts
```diff
@@ -0,0 +1,34 @@
+import { BaseTaskResult, TaskResult, ui, TableData } from '@checkup/core';
+
+export default class EmberInRepoAddonEnginesTaskResult extends BaseTaskResult
+  implements TaskResult {
+  inRepoAddons!: string[];
+  engines!: string[];
```
scalvert, line 6: We should update this to `inRepoEngines`. That's what we're looking at here.

PR #394, opened 2020-05-10
Adding how to documentation for Plugins and Tasks

scalvert:
Adds how to documentation for Plugins and Tasks within the docs/ folder.

- [Plugins](https://github.com/checkupjs/checkup/blob/add-how-to-docs/docs/how-to-create-a-plugin.md)


docs/how-to-create-a-plugin.md
```diff
@@ -0,0 +1,86 @@
...
+yarn test
+```
+
+## Configuring Your Plugin to Run in Checkup
+
+In order to use your plugin in the Checkup CLI, you need to configure it. You project needs to have a `.checkuprc` file in its root - one can be generated by running `checkup generate config`.
+
+Once your project has a configuration file, you can configure your plugin by adding the plugin name to the `plugins` array.
+
+Using our `checkup-plugin-demo` plugin above, you can configuring it either by using the fully qualified name (the name with `checkup-plugin-` preceding it), or the short name (the name without a preceding `checkup-plugin-`):
```
carakessler, line 68: I think this is confusing.. and it would be better if this only worked one way or the other. What value do you see of it working both with and without the `checkup-plugin-*` prefix?

scalvert: This is entirely inline with how eslint works. You can use either the fully qualified name, or the short name. I think there's a lot of value there.

Additionally, keeping the behavior consistent with established norms helps in users being able to have a chance at guess at how it'll work. When I use a tool and am able to guess how it'll work due to it being based on established patterns, I feel this is a delight.

scalvert: https://eslint.org/docs/user-guide/configuring#configuring-plugins

> To configure plugins inside of a configuration file, use the plugins key, which contains a list of plugin names. The eslint-plugin- prefix can be omitted from the plugin name.

scalvert: If you're ok with this, I'll just resolve this comment. Let me know!

scalvert:
Closing this in favor of https://docs.checkupjs.com/

PR #401, opened 2020-05-13
[task] Updates Outdated Dependency output to use barchart

scalvert:
New output:

Screen Shot 2020-05-13 at 3 55 35 PM


rwjblue:
Looks snazzy! IMHO, it makes the breakdown much nicer

Issue #400, opened 2020-05-11
Adding support to checkup to operate on specific paths 


carakessler:
Currently, checkup can only be run on an entire application, beginning at the root which is passed in via (the only) command line args. We need to know the root for a number of reasons, one of which being it is where checkups goes to look for the checkup config. 

Checkup needs to be able to operate on a subset of the files in a directory, primary use case being large applications with shared ownership. In order to provide detailed and targeted insights to the right people, we need to account for shared ownership of an application as a first class citizen, and provide separate checkup reports for each of the separate ownership "groups".

To accomplish this, the following changes need to be made: 
1) The paths command line arg will be used to pass in globs in the same manner [eslint does](https://eslint.org/docs/user-guide/command-line-interface). 
2) The root directory will be passed in via a `--cwd` flag, defaulting to `.`. 
3) All tasks that will operate on these globs directly, instead of assuming they operate on everything inside root. 

Implementation details of task related changes: 
Right now there are 4 types of tasks we support - the following changes will need to be made to each type to allow for this functionality. 

1) Meta tasks (provide meta info about the application such as name, days active, etc.) 
	- Meta tasks will still operate on the root
2) Tasks that read and analyze JSON  
	- Tasks that operate on package.json (dependency related tasks) will take in a flag that determines whether they operate on the root package.json, or follow the globs. **Open question - what happens for dependency tasks when multiple package.jsons are found? Tasks need to account for this.**
3) Tasks that invoke lint rules 
	- All tasks that invoke lint rules simply need to pass in the globs to the parsers, as eslint and templatelint (the only parsers we have right now), support this in a first class manner. Worth noting, any new parser that that developers intend to add in the future must have this functionality. 
4) String search/directory search
	- `file-searcher.ts` will need to be modified to account for this functionality. 

Extensions: 
1) Enable the configuration of paths to operate on in the checkup config. **Note - if there are paths specified in the config, and paths specified via command line, command line will be prioritized and paths specified in the config will be overriden by command line**

scalvert:
Thanks for getting this conversation going, @carakessler!

A couple of things:

- do we think `--root` is the right flag name, or should we use `--working-dir` or something?
- should we instead just load the top level package.json from the root dir and pass it down to tasks rather than relying on them to load it? This will make them able to operate separate from a configured path. That said, if checkup is run in a directory without a package.json, it should error (it does today)
[...]

PR #428, opened 2020-05-21
Adding pkg: PackageJson to the TaskContext and updating all tasks to use it

carakessler:
This is a part of https://github.com/checkupjs/checkup/issues/400.
Also removing getPackageJson from core, so it isnt made available to the plugins, as they shouldnt need to use it.

The one task that is still implemented questionably is `ember-in-repo-addons-engines-task` - it reads all package.jsons in a repo, I think it should likely still just operate from root, and traverse and find all package.jsons, but I will revisit that implementation when I add in support for the globs

### DIFF
--- a/packages/checkup-plugin-ember/src/tasks/ember-in-repo-addons-engines-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-in-repo-addons-engines-task.ts
@@ -8,8 +8,10 @@ import {
   TaskContext,
 } from '@checkup/core';
 import EmberInRepoAddonEnginesTaskResult from '../results/ember-in-repo-addons-engines-task-result';
-import { getPackageJson } from '@checkup/core';
+
 import { PackageJson } from 'type-fest';
+import * as path from 'path';
+import * as fs from 'fs-extra';
 
 export default class EmberInRepoAddonsEnginesTask extends FileSearcherTask implements Task {
   meta: TaskMetaData = {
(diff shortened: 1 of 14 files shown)


packages/checkup-plugin-ember/src/tasks/ember-in-repo-addons-engines-task.ts
```diff
@@ -8,8 +8,10 @@ import {
   TaskContext,
 } from '@checkup/core';
 import EmberInRepoAddonEnginesTaskResult from '../results/ember-in-repo-addons-engines-task-result';
-import { getPackageJson } from '@checkup/core';
+
 import { PackageJson } from 'type-fest';
+import * as path from 'path';
+import * as fs from 'fs-extra';
```
scalvert, line 14: I'm going to start avoiding `*` imports, mainly due to the fact that you can't see what's in use in this module easily. Can you change these to named imports?

packages/test-helpers/src/get-task-context.ts
```diff
@@ -15,15 +16,23 @@ const DEFAULT_CONFIG = {
...
+  keywords: [],
+};
+
 export function getTaskContext(
   cliArguments: {} = {},
   cliFlags: {} = DEFAULT_FLAGS,
-  config: {} = DEFAULT_CONFIG
+  config: {} = DEFAULT_CONFIG,
+  pkg: PackageJson = DEFAULT_PACKAGE_JSON
 ): TaskContext {
```
scalvert, line 30: This signature is getting a bit unwieldy. We should consider a refactor where we pass in an object and just `Object.assign` or spread it to override.

scalvert: We don't need to do this now, but can do it in a subsequent PR!

carakessler: totally agree - and its only going to grow when i add paths. was planning on cleaning up after tthat!

scalvert: 👍

PR #430, opened 2020-05-23
Adding paths to TaskContext and updating all tasks to use it

carakessler:
This completes the work outlined in this issue https://github.com/checkupjs/checkup/issues/400

Essentially, adding `paths` to the `TaskContext` and updating all existing tasks and tests to operate on globs (if they are passed in) instead of the entire app.


packages/cli/src/commands/run.ts
```diff
@@ -157,6 +161,7 @@ export default class RunCommand extends Command {
       parsers: getRegisteredParsers(),
       config: this.checkupConfig,
       pkg: getPackageJson(this.runFlags.cwd),
+      paths: getFilePaths(this.runFlags.cwd, this.runArgs as string[]),
```
carakessler, line 164: remove RunArgs type, it is just a string[] now

packages/core/package.json
```diff
@@ -6,13 +6,16 @@
...
     "ajv": "^6.12.2",
+    "@types/micromatch": "^4.0.1",
     "chalk": "^4.0.0",
     "cli-ux": "^5.4.4",
     "debug": "^4.1.1",
     "ember-template-lint": "^2.8.0",
     "eslint": "^6.8.0",
     "fs-extra": "^9.0.0",
     "globby": "^11.0.0",
+    "io-ts": "^2.2.2",
```
scalvert, line 17: What do you need `io-ts` for? It was removed as part of my config change.

packages/core/src/searchers/file-searcher.ts
```diff
@@ -1,109 +1,54 @@
...
-    await Promise.all(
-      files.map((file) => {
-        return fs.promises
-          .readFile(path.join(path.resolve(this.baseDirectory), file), 'utf8')
-          .then((fileString: string) => {
-            for (let searchPatternName in this.searchPatterns) {
-              stringResults.counts[searchPatternName] += this._getStringSearchCount(
-                fileString,
-                this.searchPatterns[searchPatternName]
+export async function findStrings(
```
scalvert, line 6: I'm not a fan of having the type as part of the name. Can we just call this `findInFiles`, or `findContents`?

packages/test-helpers/src/get-task-context.ts
```diff
@@ -22,17 +22,19 @@ const DEFAULT_PACKAGE_JSON = {
...
-  cliFlags: {} = DEFAULT_FLAGS,
-  config: {} = DEFAULT_CONFIG,
-  pkg: PackageJson = DEFAULT_PACKAGE_JSON
-): TaskContext {
+export function getTaskContext({
+  cliArguments = [] as [],
+  cliFlags = DEFAULT_FLAGS as {},
+  config = DEFAULT_CONFIG as {},
+  pkg = DEFAULT_PACKAGE_JSON as PackageJson,
+  paths = [] as string[],
```
scalvert, line 30: Why do you need to safe cast these?

packages/test-helpers/src/filter-result-data.ts
```diff
@@ -0,0 +1,9 @@
+import { TaskItemData } from '@checkup/core';
+
+// the paths in the json result are dynamic, removing them for checking against the snapshot
+export function filterTaskItemDataJson(types: TaskItemData[]) {
```
scalvert, line 4: Isn't this really `clearFilePaths`?

packages/test-helpers/src/checkup-fixturify-project.ts
```diff
@@ -76,4 +78,23 @@ export default class CheckupFixturifyProject extends Project {
 
     this.pkg = packageJsonContent;
   }
+
+  getFilePaths(): string[] {
```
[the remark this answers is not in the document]
carakessler: sure

packages/test-helpers/src/get-task-context.ts
carakessler: without it for `cliFlags` and `config`, even though we are merging the passed in flags with the defaults, it throws an error that the `cliFlags` being passed in dont have all of the required fields. 

```
Type '{ cwd: string; }' is missing the following properties from type 
'{ version: undefined; help: undefined; config: undefined; 
cwd: string; task: undefined; format: string; outputFile: string; }': 
version, help, config, task, and 2 more.ts(2740)

get-task-context.d.ts(5, 5): The expected type comes from property 
'cliFlags' which is declared here on type '{ cliArguments?: never[]; 
cliFlags?: { version: undefined; help: undefined; config: undefined; 
cwd: string; task: undefined; format: string; outputFile: string; }; config?: {}; 
pkg?: PackageJson; paths?: string[]; }'

```

scalvert: You can make the type `Partial`.

```ts
Partial<RunFlags>
```

I think part of the issue here is that you didn't type the object's properties.

PR #444, opened 2020-05-27
Adding LOC count task

carakessler:
Adding task that counts the number of LOC per file type in an app.

Note: the sloc module also handles counting TODOs, so I will be removing the todo task in a subsequent PR


packages/cli/src/tasks/lines-of-code-task.ts
```diff
@@ -0,0 +1,107 @@
...
+const sloc = require('sloc');
+const micromatch = require('micromatch');
+
+/*
+ * note: these extensions must be supported here https://github.com/flosse/sloc/blob/731fbea00799a45a6068c4aaa1d6b7f67500615e/src/sloc.coffee#L264
+ * for the analysis on comments/empty lines/etc to be correct
+ */
+const FILE_EXTENSIONS_SUPPORTED = ['js', 'hbs', 'scss', 'css', 'json', 'md'];
+
+interface GroupedFiles {
```
scalvert, line 22: Small nit, but I think it's useful to have this ordering be consistent in files:

imports
type definitions
variable declarations
functions, classes, etc

By keeping the type defs separate, at the top, they're easily distinguishable from code that's relevant to the running of the module. Thoughts?

packages/cli/src/tasks/lines-of-code-task.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const fs = require('fs');
+const sloc = require('sloc');
+const micromatch = require('micromatch');
+
+/*
+ * note: these extensions must be supported here https://github.com/flosse/sloc/blob/731fbea00799a45a6068c4aaa1d6b7f67500615e/src/sloc.coffee#L264
+ * for the analysis on comments/empty lines/etc to be correct
+ */
+const FILE_EXTENSIONS_SUPPORTED = ['js', 'hbs', 'scss', 'css', 'json', 'md'];
```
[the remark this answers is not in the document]
carakessler: so we can go one of two ways here... we can have it be configurable, which is more performant, but requires developers telling us what types of files are in their repo, which we can glean ourselves. or we can add some code (maybe as a part of `TaskContext`) that lists all the extensions in the repo and allows this thing to be totally dynamic. thoughts? this hardcoded list is definitely not meant to be final solution, just wanted to get PR up so we could have this conversation.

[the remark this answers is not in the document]
carakessler: Discussed offline - we are going to get all file extensions in the repo

packages/cli/src/results/lines-of-code-task-result.ts
```diff
@@ -0,0 +1,82 @@
...
+        comment: 0,
+        empty: 0,
+        mixed: 0,
+        single: 0,
+        source: 0,
+      }
+    );
+
+    return {
+      results: [reducedResults],
```
scalvert, line 77: Why does this need to be wrapped in an array?

packages/cli/src/tasks/lines-of-code-task.ts
```diff
@@ -0,0 +1,120 @@
...
+  constructor(pluginName: string, context: TaskContext) {
+    super(pluginName, context);
+
+    // get all file exensions in the repo to group the files
+    const fileExensionsInRepo = new Set();
+    this.context.paths.forEach((path) => {
+      fileExensionsInRepo.add(path.split('.')[1]);
+    });
+
+    this.filesGroupedByType = ([...fileExensionsInRepo] as string[]).map((ext) => {
```
scalvert, line 75: I _think_ if you use a `Set<string>` above you won't have to cast this here.

carakessler: truth thanks!

packages/cli/src/results/lines-of-code-task-result.ts
carakessler: unwrapped that present

PR #445, opened 2020-05-27
Adding task that counts the number of eslint overrides

carakessler:
Adding task that counts the number of eslint overrides

### DIFF
--- a/packages/cli/src/tasks/eslint-disable-task.ts
+++ b/packages/cli/src/tasks/eslint-disable-task.ts
@@ -1,0 +1,74 @@
+import {
+  TaskResult,
+  Task,
+  Category,
+  Priority,
+  TaskMetaData,
+  BaseTask,
+  TaskItemData,
+} from '@checkup/core';
+import EslintDisableTaskResult from '../results/eslint-disable-task-result';
+
+const fs = require('fs');
+const micromatch = require('micromatch');
+const recast = require('recast');
+const babel = require('@babel/parser');
+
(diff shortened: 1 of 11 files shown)


packages/cli/src/tasks/eslint-disable-task.ts
```diff
@@ -0,0 +1,73 @@
...
+    errors: [] as string[],
+    results: [] as TaskItemData[],
+  };
+
+  await Promise.all(
+    filePaths.map((file) => {
+      return fs.promises
+        .readFile(file, 'utf8')
+        .then((fileString: string) => {
+          let objAst = recast.parse(fileString, {
```
scalvert, line 50: Let's drop the Hungarian notation in favor of `ast`.

packages/cli/src/tasks/eslint-disable-task.ts
```diff
@@ -0,0 +1,73 @@
...
+            parser: {
+              parse: babel.parse,
+            },
+          });
+          let numMatches = 0;
```
scalvert, line 55: Could we declare vars at the top of the scope? This keeps the code easier to read. Also, it wouldn't hurt to give this code a bit of breathing room in the form of blank lines so it's easier to parse visually.

scalvert:
Meta question: should we start to standardize _where_ we test tasks? I mean should we simply not test tasks via the CLI in `run-test`, but rather just use unit tests on the tasks themselves in their individual files? I think we're duplicating some tests here, which may make maintenance a bit harder.

PR #448, opened 2020-05-28
Gracefully handle errors in the CLI

scalvert:
Adding more useful errors to the CLI. 

- Adds new `CheckupError` class that allows providing not only a `message` property, but a `callToAction`
- Updates CLI error paths to use `CheckupError`
- Changes `TaskList` and `MetaTaskList` to catch and store errors for running tasks, which allows for graceful error handling for errors during task execution.

Examples:

Error when no `.checkuprc` present:

Screen Shot 2020-05-28 at 10 58 09 AM

Errors for invalid `.checkuprc`:

Screen Shot 2020-05-28 at 10 58 42 AM

Screen Shot 2020-05-28 at 10 59 14 AM

Error when no single task found:

Screen Shot 2020-05-28 at 11 00 36 AM


carakessler:
Discussed offline - there will be a followup PR that collects the nonterminal errors from each task, and propagates them up. For example, all tasks that read files collect array of errors of files that it failed to read. This may not cause the whole task to fail, but is worth surfacing to user - i.e here are all your results except we couldnt read `lib/foo.js` so it is not encompassed in the results

packages/core/src/errors/checkup-error.ts
```diff
@@ -0,0 +1,29 @@
+import * as Wrap from 'wrap-ansi';
+
+import { CLIError } from '@oclif/errors';
+import { red } from 'chalk';
+
+export default class CheckupError extends CLIError {
+  constructor(
+    error: string | Error,
+    public callToAction: string = '',
```
carakessler, line 9: do we want the CTA to be optional? an error without a CTA seems questionable...

scalvert: Maybe not :)

I can make it required. If people don't want to set the CTA, they can use the `Error` class instead.

PR #453, opened 2020-05-29
Adding task that runs eslint as configured in the app checkup is being run on

carakessler:
This run eslint as configured in the app checkup is being run on, and summarizes the results.

### DIFF
--- a/packages/checkup-plugin-javascript/__tests__/eslint-summary-task-test.ts
+++ b/packages/checkup-plugin-javascript/__tests__/eslint-summary-task-test.ts
@@ -1,4 +1,5 @@
 import { CheckupProject, stdout, getTaskContext } from '@checkup/test-helpers';
+import { ESLintReport } from '@checkup/core';
 import {
   EslintSummaryTask,
   readEslintConfig,
@@ -6,7 +7,6 @@ import {
 } from '../src/tasks/eslint-summary-task';
 import EslintSummaryTaskResult from '../src/results/eslint-summary-task-result';
 import { PackageJson } from 'type-fest';
-import { CLIEngine } from 'eslint';
 
 describe('eslint-summary-task', () => {
   let project: CheckupProject;
@@ -127,7 +127,7 @@ describe('readEslintConfig', () => {
   });
 });
 
-function filterResultFilePath(report: CLIEngine.LintReport) {
+function filterResultFilePath(report: ESLintReport) {
   let filteredResults = report.results.map((result) => {
     result.filePath = '';
   });
--- a/packages/checkup-plugin-javascript/__tests__/eslint-summary-task-test.ts
+++ b/packages/checkup-plugin-javascript/__tests__/eslint-summary-task-test.ts
@@ -1,0 +1,135 @@
+import { CheckupProject, stdout, getTaskContext } from '@checkup/test-helpers';
+import {
+  EslintSummaryTask,
+  readEslintConfig,
+  ACCEPTED_ESLINT_CONFIG_FILES,
+} from '../src/tasks/eslint-summary-task';
+import EslintSummaryTaskResult from '../src/results/eslint-summary-task-result';
+import { PackageJson } from 'type-fest';
+import { CLIEngine } from 'eslint';
+
+describe('eslint-summary-task', () => {
+  let project: CheckupProject;
+
+  beforeEach(() => {
+    project = new CheckupProject('checkup-app', '0.0.0');
+    project.files['foo.js'] = `function foo() {
+        var whatever = 'ESLint'
+        return {
+          name: whatever
+        };
+      }`;
+    project.files['.eslintrc.js'] = `module.exports = {
+        rules: {
+            semi: 'error',
+            'no-var': 'warn'
+        }
+      };`;
+    project.writeSync();
+    project.gitInit();
+    project.install();
+  });
+
+  afterEach(() => {
+    project.dispose();
+  });
+
+  it('it summarizes eslint and outputs to console', async () => {
+    const result = await new EslintSummaryTask(
+      'meta',
+      getTaskContext({
+        cliFlags: { cwd: project.baseDir },
+        pkg: project.pkg,
+        paths: project.filePaths,
+      })
+    ).run();
+    const taskResult = <EslintSummaryTaskResult>result;
+
+    taskResult.toConsole();
+
+    expect(stdout()).toMatchInlineSnapshot(`
+      "=== Eslint Summary ===
+
+      Error count: 1
+      Warning count: 1
+
+      === Errors
+
+      Rule name Failures             
+      semi      1 errors (1 fixable) 
+
+      === Warnings
+
+      Rule name Failures               
+      no-var    1 warnings (1 fixable) 
+
+      "
+    `);
+  });
+
+  it('it summarizes eslint and outputs to JSON', async () => {
+    const result = await new EslintSummaryTask(
+      'meta',
+      getTaskContext({
+        cliFlags: { cwd: project.baseDir },
+        pkg: project.pkg,
+        paths: project.filePaths,
+      })
+    ).run();
+    const taskResult = <EslintSummaryTaskResult>result;
+
+    expect(filterResultFilePath(taskResult.toJson().result.esLintReport)).toMatchSnapshot();
+  });
+});
+
+describe('readEslintConfig', () => {
+  const eslintConfigJson = 'airbnb';
+  const pkg: PackageJson = {
+    name: 'foo-project',
+    version: '0.0.0',
+    keywords: [],
+    eslintConfig: eslintConfigJson,
+  };
+
+  ACCEPTED_ESLINT_CONFIG_FILES.forEach((file) => {
+    it(`it returns ${file}, even if there is also a JSON config (as .eslintrc* are prioritized over config in package.json)`, () => {
+      let expectedConfig = 'foo';
+
+      let project = new CheckupProject('foo-project');
+      project.files[file] = expectedConfig;
+      project.writeSync();
+
+      const eslintConfig = readEslintConfig(project.filePaths, project.baseDir, pkg);
+
+      expect(eslintConfig).toEqual(expectedConfig);
+      project.dispose();
+    });
+  });
+
+  it('it reads config from package.json, only if there is no .eslintrc*', () => {
+    const eslintConfig = readEslintConfig([], '.', pkg);
+
+    expect(eslintConfig).toEqual(eslintConfigJson);
+  });
+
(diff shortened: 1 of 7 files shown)


packages/checkup-plugin-javascript/__tests__/eslint-summary-task-test.ts
```diff
@@ -0,0 +1,135 @@
...
+    project.writeSync();
+
+    const eslintConfig = readEslintConfig(project.filePaths, project.baseDir, pkg);
+
+    expect(eslintConfig).toEqual(expectedConfig);
+    project.dispose();
+  });
+});
+
+function filterResultFilePath(report: CLIEngine.LintReport) {
```
scalvert, line 130: We have a specific type we export from core for this so we can keep the type defs centralized.

```ts
export type ESLintReport = CLIEngine.LintReport;
```

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
```diff
@@ -0,0 +1,113 @@
...
+  failureCount: number;
+  fixableFailureCount: number;
+};
+
+type EslintDataForConsole = {
+  errorCount: number;
+  warningCount: number;
+  errorList: { rule: string; failures: string }[];
+  warningList: { rule: string; failures: string }[];
+};
```
scalvert, line 15: Why is this type only for console?

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export default class EslintSummaryTaskResult extends BaseTaskResult implements TaskResult {
+  esLintReport!: ESLintReport;
+
+  toConsole() {
+    let sortedEslintReport = transformEslintReport(this.esLintReport);
```
scalvert, line 21: I'd love it if we could consolidate these transform functions around a pattern that ppl could follow vs. ad hoc functions we make. Even having a function interface would at least allow authors to follow a well trodden path. Nothing necessarily we should do right now, but we should definitely be thinking about it.

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
```diff
@@ -0,0 +1,113 @@
...
+  let errors: Record<string, EslintFailureCounts> = {};
+  let warnings: Record<string, EslintFailureCounts> = {};
+
+  // we want to strip these from total counts (generally they are just a warning that a file was ignored)
+  let warningsWithoutRulesIds = 0;
+
+  esLintReport.results.forEach((result) => {
+    if (result.messages.length > 0) {
+      result.messages.forEach((msg) => {
+        // severity 1 is a warning
```
scalvert, line 62: Should you create an enum so this reads better without the need for a comment?

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
```diff
@@ -0,0 +1,113 @@
...
+
+function createFailureList(record: Record<string, EslintFailureCounts>, failureType: string) {
+  let chalkColor: Function = failureType === 'errors' ? chalk.red : chalk.yellow;
+
+  return Object.entries(record).map((arr) => {
+    return {
+      rule: arr[0],
+      failures: `${chalkColor(arr[1].failureCount + arr[1].fixableFailureCount)} ${failureType} (${
+        arr[1].fixableFailureCount
+      } fixable)`,
```
scalvert, line 110: Can you store the value of these array indexes so we can make more sense of what the values are?

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @description In prioritized order as specified by https://eslint.org/docs/user-guide/configuring
+ */
+export const ACCEPTED_ESLINT_CONFIG_FILES = [
+  '.eslintrc.js',
+  '.eslintrc.cjs',
+  '.eslintrc.yaml',
+  '.eslintrc.yml',
+  '.eslintrc.json',
+  '.eslintrc',
+];
```
scalvert, line 30: Was there nothing we could borrow from eslint to read configs?

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
```diff
@@ -0,0 +1,95 @@
...
+  meta: TaskMetaData = {
+    taskName: 'eslint-summary',
+    friendlyTaskName: 'Eslint Summary',
+    taskClassification: {
+      category: Category.Insights,
+      priority: Priority.Medium,
+    },
+  };
+
+  private eslintParser: Parser<ESLintReport>;
```
scalvert, line 42: We should try to underscore private stuff, or use private fields.

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
```diff
@@ -0,0 +1,95 @@
...
+    super(pluginName, context);
+
+    let createEslintParser = this.context.parsers.get('eslint')!;
+
+    let eslintConfig = readEslintConfig(
+      this.context.paths,
+      this.context.cliFlags.cwd,
+      this.context.pkg
+    );
+    this.eslintParser = createEslintParser(eslintConfig as ESLintOptions);
```
scalvert, line 55: Why not give the var above a type, then you won't have to cast it.

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.jsPaths = this.context.paths.filterByGlob('**/*.js');
+  }
+
+  async run(): Promise<TaskResult> {
+    let result: EslintSummaryTaskResult = new EslintSummaryTaskResult(this.meta);
+
+    let esLintReport = await this.runEsLint();
+    this.debug('ESLint Report', esLintReport);
+    result.esLintReport = esLintReport;
+    return result;
```
scalvert, line 66: Nit: breathing room :)

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (eslintConfigFile) {
+    return eslintConfigFile;
+  } else if (pkg.eslintConfig !== null) {
+    return pkg.eslintConfig as string;
+  } else {
+    throw new Error(
+      'No eslint config found in root (in the form of .eslintrc.* or as an eslintConfig field in package.json)'
+    );
+  }
+}
```
scalvert, line 95: Is there nothing we can use from eslint itself to read configs?

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
carakessler: yea its definitely something that has been top of mind to me as i have written a bunch of tasks this week.. i honestly have seen too many patterns emerge from these various data transformations, but they are definitely the brunt of the work when it comes to writing a task rn

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
carakessler: the only thing i could find exported was `getConfigForFile` which obviously doesnt work for our use case

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
carakessler: the only thing i could find exported was getConfigForFile which obviously doesnt work for our use case

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
carakessler: we can give the var above a type, and i can cast in the `readEslintConfig` function and have that return `ESLintOptions`, but reading the file is going to return a string, and at some point i need to cast that to `ESLintOptions`, no?

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
scalvert: Yep. This is why I think it's useful for us to start thinking about schemas for the JSON we're producing in our tasks. Right now we're not adhering to any sort of schema, but rather just outputting what we think we want. We should probably do a future pass over all the output and figure out if we can provide some consistency.

packages/checkup-plugin-javascript/src/tasks/eslint-summary-task.ts
scalvert: Ya I think if you have a function, like above, that reads an eslint config and returns a structured type, that function should just return the type itself. That way consumers won't have to guess at the structure - the type will dictate that.

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
carakessler: Adding to meeting agenda

packages/checkup-plugin-javascript/src/results/eslint-summary-task-result.ts
carakessler: i am using the types provided by eslint, you can click into this and see that the severity is `type Severity = 0 | 1 | 2;`, but it doesnt say anywhere in the eslint types that 1 corresponds with warning :( read that in the docs

scalvert: I think we define that type, don't we? If so, we can define 0, 1, 2 to mean the corresponding `off`, `warning`, `error` in eslint, no?

carakessler: no, we are getting that type from ESlint

scalvert: I was thinking of this https://github.com/checkupjs/checkup/blob/master/packages/core/src/types/ember-template-lint.ts#L1

scalvert: Also, this file isn't importing/using any severity type. What type are you talking about? My proposal is to define `Severity`, or use the type above, and use that for comparison.

```
if (msg.severity === Severity.warning) {
  // ...
}
```

scalvert: Let's just leave this for now.

PR #466, opened 2020-06-01
Adding field excludePaths to config, so consumers of checkup can add globs to be ignored

carakessler:


packages/cli/__tests__/commands/run-test.ts
```diff
@@ -131,5 +131,38 @@ describe('@checkup/cli', () => {
...
+          foo: {
+            'index.hbs': '{{!-- i should todo: write code --}}',
+          },
+        });
+
+        anotherProject.writeSync();
+
+        await runCommand(['run', '--cwd', anotherProject.baseDir]);
+        let filteredRun = stdout();
+        expect(filteredRun).toMatchSnapshot();
```
scalvert, line 150: Can we use inline snapshots here so it's easier to see what the expected is at this point?

packages/cli/__tests__/commands/run-test.ts
```diff
@@ -131,5 +131,38 @@ describe('@checkup/cli', () => {
...
+
+        clearStdout();
+
+        anotherProject.addCheckupConfig({ pathsToIgnore: [] });
+        anotherProject.writeSync();
+
+        await runCommand(['run', '--cwd', anotherProject.baseDir]);
+        let unFilteredRun = stdout();
+        expect(unFilteredRun).toMatchSnapshot();
```
scalvert, line 159: Same inline snapshot comment.

packages/core/src/types/config.ts
```diff
@@ -1,5 +1,6 @@
 export type TaskConfig = 'on' | 'off' | ['on' | 'off', unknown];
 export type CheckupConfig = {
+  pathsToIgnore: string[];
```
scalvert, line 3: Can we rename this to `excludePaths`?

scalvert:
Are we planning on adding a new command line flag to provide another way to exclude paths?

packages/cli/__tests__/commands/run-test.ts
carakessler: Hmm my thoughts here are these because these snapshots are long and ugly, i think it makes sense to have them in the snapshot files. that is why i added an additional assertion in this type of test ensuring that the two snapshots generated are not equal (which is really the main thing that matters for the test), so people don't have to read through the actual snapshots

carakessler:
@scalvert added support for command line flag - updating PR now

packages/cli/__tests__/commands/run-test.ts
scalvert: That makes sense. Thanks.

PR #467, opened 2020-06-01
Creating FilePathsArray class that extends array and offers filtering functionality

carakessler:
Many of the tasks were taking the paths string[] from TaskContext and using micromatch to filter by globs. Now, the paths object in TaskContext is of type FilePaths which extends array and adds function filterByGlob, to make it easier for task authors to filter the file paths array. This obscures the implementation detail of using micromatch to authors.

### DIFF
--- a/packages/core/src/utils/get-paths.ts
+++ b/packages/core/src/utils/get-paths.ts
@@ -1,5 +1,6 @@
 import { join, resolve } from 'path';
 import { existsSync } from 'fs';
+import { FilePathsArray } from './file-paths-array';
 
 const micromatch = require('micromatch');
 const isValidGlob = require('is-valid-glob');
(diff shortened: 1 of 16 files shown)


packages/core/src/utils/get-paths.ts
```diff
@@ -17,10 +17,16 @@ const PATHS_TO_IGNORE = [
   '.git',
 ];
 
+export class FilePaths extends Array<string> {
```
scalvert, line 20: We should name this `FilePathsArray`, as it extends from a concrete type and it'll be a more clear hint to users that this is a derived type from Array.

carakessler: ah good call - updating

packages/core/src/utils/get-paths.ts
```diff
@@ -17,10 +17,16 @@ const PATHS_TO_IGNORE = [
   '.git',
 ];
 
+export class FilePathsArray extends Array<string> {
```
scalvert, line 20: Can we extract this to its own module, jsdoc comment it, and add tests for this class itself separate from the `get-paths` tests?

PR #474, opened 2020-06-04
(bug fix) Restricting file types for LOC class

carakessler:
Previously, we were taking all "extensions" and using them to call sloc. Extensions were defined as a string after a period. However, when running this on a real app, there was a ton of random garbage getting thrown in there. I decided to just use the exensions supported by sloc, and just add
two additional (json and md). This fits our current use case, and we can re-evaluate (also re-evaluate using this subpar library) if things change and it no longer fits our use case.

### DIFF
--- a/packages/cli/src/tasks/lines-of-code-task.ts
+++ b/packages/cli/src/tasks/lines-of-code-task.ts
@@ -48,7 +48,8 @@ export interface FileResults {
  * note: these extensions must be supported here https://github.com/flosse/sloc/blob/731fbea00799a45a6068c4aaa1d6b7f67500615e/src/sloc.coffee#L264
  * for the analysis on comments/empty lines/etc to be correct
  */
-const FILE_EXTENSIONS_SUPPORTED = sloc.extensions;
+const FILE_EXTENSIONS_SUPPORTED_BY_SLOC = new Set(sloc.extensions);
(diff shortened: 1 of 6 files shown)


packages/cli/src/tasks/lines-of-code-task.ts
```diff
@@ -67,16 +68,24 @@ export default class LinesOfCodeTask extends BaseTask implements Task {
...
+      ...FILE_EXTENSIONS_SUPPORTED_BY_SLOC,
+    ]);
+
     this.context.paths.forEach((path) => {
-      fileExensionsInRepo.add(path.split('.')[1]);
+      let split = path.split('.');
+      if (fileExtensionsToLookFor.has(split[1])) {
+        fileExensionsInRepo.add(split[1]);
+      }
     });
```
scalvert, line 81: I think you could just use an array filter here, no?

carakessler: i think eslint-plugin-unicorn had some new "prefer-set" rule or something like that which converted this to Set.has

scalvert: Sorry I mean couldn't we just filter out the paths based on `fileExtensionsToLookFor` and just return into `fileExtensionsInRepo`?

carakessler: ah so just add all file extensions in repo to the set, and then filter at the end. ya that prob more performant

PR #476, opened 2020-06-05
TaskClassification refactor

carakessler:
This PR accomplishes the following:
1) Removes "Priority" as a concept from checkup (will be replaced by Action Items)
2) Migrate what we currently call "Category" to be called "TaskType"
3) Category is now a string NOT an enum

The tasks are sorted first by Category, then by TaskType.

Examples of Categories: 'ember', 'testing', 'linting', 'dependencies'.
This is how we will group like tasks from different plugins together


packages/checkup-plugin-ember-octane/src/tasks/octane-migration-status-task.ts
```diff
@@ -18,8 +17,8 @@ export default class OctaneMigrationStatusTask extends BaseTask implements Task
     taskName: 'octane-migration-status',
     friendlyTaskName: 'Ember Octane Migration Status',
     taskClassification: {
-      category: Category.Migrations,
-      priority: Priority.High,
+      taskType: TaskType.Migrations,
```
scalvert, line 20: Should we just make this a `string` too, for maximum flexibility?

packages/cli/__tests__/generators/__snapshots__/generate-task-test.ts.snap
```diff
@@ -1,16 +1,16 @@
...
 
 export default class MyFooTask extends BaseTask {
   meta = {
     taskName: 'my-foo',
     friendlyTaskName: 'My Foo',
     taskClassification: {
-      category: Category.Insights,
-      priority: Priority.Low,
+      taskType: TaskType.Insights,
+      category: '' ,
```
scalvert, line 13: I think we should disallow empty, no?

packages/cli/src/tasks/lines-of-code-task.ts
```diff
@@ -58,8 +50,8 @@ export default class LinesOfCodeTask extends BaseTask implements Task {
     taskName: 'lines-of-code',
     friendlyTaskName: 'Lines of Code',
     taskClassification: {
-      category: Category.Insights,
-      priority: Priority.Low,
+      taskType: TaskType.Insights,
+      category: 'meta', // TODO: change this to a meta task
```
scalvert, line 54: Or put in js plugin?

carakessler: LOC will work outside the scope of JS projects... i feel like this is more meta info about a repo, similar to the other data collected by meta tasks. thoughts?

packages/cli/__tests__/generators/__snapshots__/generate-task-test.ts.snap
carakessler: hmmm should we? the empty ones will all be collected together at the end, which i guess would be weird. trying to think about whether or not having a category in mind and ready should be a blocker to creating a task, and I am leaning towards not, but it could yield unruly and unrelated results being gathered at the end.. more thinking out loud here. or quietly typing my thoughts #WFHforever. ya okay im on board, we shouldnt allow for empty, i will add that in :)

packages/checkup-plugin-ember-octane/src/tasks/octane-migration-status-task.ts
carakessler: ehh maybe? open to the idea, would love to talk about it. but if so, that can be a separate refactor anyway, so resolving this but adding the topic to our meeting notes

packages/checkup-plugin-ember-octane/__tests__/results/__snapshots__/octane-migration-status-tast-result-test.ts.snap
```diff
@@ -5,8 +5,8 @@ Object {
   "meta": Object {
     "friendlyTaskName": "Ember Octane Migration Status",
     "taskClassification": Object {
-      "category": "migrations",
-      "priority": "high",
+      "category": "ember",
+      "taskType": "migrations",
```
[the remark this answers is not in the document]
carakessler: sure

packages/cli/src/tasks/lines-of-code-task.ts
scalvert: Seems good for now 👍

packages/cli/__tests__/task-list-test.ts
```diff
@@ -25,7 +25,18 @@ describe('TaskList', () => {
...
+  });
+
+  it('registerTask fails if a task doesnt have a category set', () => {
+    let taskList = new TaskList();
+    let taskWithoutCategory = new TaskWithoutCategory(getTaskContext());
+
+    expect(() => {
+      taskList.registerTask(taskWithoutCategory);
+    }).toThrow(
+      `Task category can not be empty. Please add a category to ${taskWithoutCategory.meta.friendlyTaskName}.`
```
scalvert, line 38: I think we should use `taskName` here, as the friendly name is really only used for output. The `taskName` is more recognizable, in that you'd be able to format this like `${taskName}-task`, and that would essentially be the filename.

carakessler: sg!

PR #486, opened 2020-06-08
Render the task category in stdout

carakessler:
In a previous PR, I added the concept of a string `category` to checkup, and sorted the results by category. In this PR, I actually render the category to stdout for each group as a header (bold, with the `===` before and after the string), and changed the task headers to not render the `===`, making them more sort of subheaders.

Attached before and after to help show the UI changes.

### DIFF
--- a/packages/checkup-plugin-ember-octane/__tests__/results/__snapshots__/octane-migration-status-tast-result-test.ts.snap
+++ b/packages/checkup-plugin-ember-octane/__tests__/results/__snapshots__/octane-migration-status-tast-result-test.ts.snap
@@ -2014,7 +2014,7 @@ Object {
 `;
 
 exports[`octane-migration-status-task-result console output simple console output 1`] = `
-"=== Ember Octane Migration Status ===
+"Ember Octane Migration Status
 
 Octane Violations: 35
 
--- a/packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
+++ b/packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
@@ -6,7 +6,7 @@ Checkup report generated for checkup-app v0.0.0
 
 This project is 0 days old, with 0 days active days, 0 commits and 0 files.
 
-=== Meta
+=== Project Info
 
 Lines of Code
 
@@ -27,7 +27,7 @@ Checkup report generated for checkup-app v0.0.0
 
 This project is 0 days old, with 0 days active days, 0 commits and 0 files.
 
-=== Meta
+=== Project Info
 
 Lines of Code
 
@@ -48,7 +48,7 @@ Checkup report generated for checkup-app v0.0.0
 
 This project is 0 days old, with 0 days active days, 0 commits and 0 files.
 
-=== Meta
+=== Project Info
 
 Lines of Code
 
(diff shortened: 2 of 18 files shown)


carakessler:
before
before change

carakessler:
after
after change

carakessler:
There are likely more changes we can make to the individual tasks to clean this up more (for example - `Ember Types` is now under the `Ember` category, so the word `Ember` can likely be removed), just wanted to get feedback on the direction before we iteratively clean up further

packages/core/src/types/tasks.ts
```diff
@@ -31,6 +31,7 @@ export interface TaskContext {
 }
 
 export interface TaskResult {
+  meta: TaskMetaData;
```
carakessler, line 34: we need the `TaskMetaData` here to get the task category in the stdout. This data was always there already, because all `*-task-result`s `extends BaseTaskResult implements TaskResult`, so no changes to the actual tasks were made, but meta was added to this interface.

packages/checkup-plugin-ember-octane/__tests__/results/__snapshots__/octane-migration-status-tast-result-test.ts.snap
```diff
@@ -2014,7 +2014,7 @@ Object {
 `;
 
 exports[`octane-migration-status-task-result console output simple console output 1`] = `
-"=== Ember Octane Migration Status ===
+"Ember Octane Migration Status
```
scalvert, line 2017: This seems fine, but the idea I had with subheadings was that they were using indentation to indicate they were sub-headings. It's a bit odd to see a main heading be further indented than a sub - you'd expect it to be the other way around, so you can visually distinguish between a main heading and a sub. Thoughts?

packages/cli/__tests__/commands/__snapshots__/run-test.ts.snap
```diff
@@ -6,12 +6,15 @@ Checkup report generated for checkup-app v0.0.0
 
 This project is 0 days old, with 0 days active days, 0 commits and 0 files.
 
-=== Lines of Code ===
+=== Meta
```
scalvert, line 9: I'm not super fond of `Meta`, as it means nothing to users. It's an internal term. I'd prefer we just don't have any title for Meta tasks.

packages/cli/src/reporters.ts
```diff
@@ -53,7 +53,20 @@ export function getReporter(
...
           .forEach((taskResult) => taskResult.toConsole());
 
-        pluginTaskResults.forEach((taskResult) => taskResult.toConsole());
+        let currentCategory = '';
+
+        pluginTaskResults.forEach((taskResult) => {
+          let taskCategory = taskResult.meta.taskClassification.category;
+
+          if (taskCategory !== currentCategory) {
+            ui.categoryHeader(_toSentenceCase(taskCategory));
```
scalvert, line 62: We should likely just add the sentence case call inside `categoryHeader`.

packages/cli/src/reporters.ts
```diff
@@ -81,3 +94,9 @@ export function getReporter(
       return async () => {};
   }
 }
+
+export function _toSentenceCase(str: string) {
```
scalvert, line 98: Lodash has a function `_.startCase` that does this.

packages/core/src/types/tasks.ts
scalvert: Yep, seems fine.

Issue #475, opened 2020-06-04
Add "Action Items" as a construct to checkup


carakessler:
**Summary**

At the bottom of the checkup report in console (also be pulled into a separate section for JSON results) we want to show a list of `Action Items` so a dev running checkup can know what requires immediate attention. Tasks can be highlighted as urgently actionable, phrased in a way that makes it clear to a dev what they are meant to do to fix. Once an `Action Item` is complete, it will no longer show up in the `Action Items` list (unlike tasks, which are purely informational, not prescriptive, and will show up with zero results). For example, if you have zero skipped tests, that result would show up. But if there was an `Action Item` configured to reduce number of skipped tests, and you had zero, it wouldn't show up  as an `Action Item` for you - since you have already done it. 

We want to make it possible for both task authors and those configuring checkup to add items to the `Action Items` list.

**Examples**

Examples of tasks that would benefit from having `Action Items` configured by authors: 
1) Using yarn 
2) Having proper linting set up 
3) Using volta 

^^^^ The theme emerging here is maybe "Project Prerequisites" or things related to the setup/toolchain of your app. 

Examples of why those configuring checkup may want to set up custom `Action Items`: 
1) We are migrating away from pattern _foo_ and it is blocking an important upgrade. We need everyone to work on prioritizing moving off of _foo_
2) There is a cleanup initiative within an app to remove `lint-disables`, we may want to configure a threshold such that prioritizing removing `lint-disables` would come up as an `Action Item` - so if your app has more than _x_ `lint-disables`, it would show up as an `Action Item` 
3) For a task that breaks down the testing pyramid, we may have there be a conditional `Action Item`, such that if you have fewer than _x_ percent of your tests as unit tests, show an `Action Item` to prioritize refactoring your test suite

^^^^^ theme emerging here is problematic patterns (potentially blocking other work) and overall health of an area of the app.

**Requirements** 
1) `Action Items` are conditional (by default in a boolean manner - i.e number of problems > 0, but can be configured more explicitly as defined by the task)
2) `Action Items` can be turned on by default in a task by the task author 
3) Tasks that are turned on as `Action Items` by default can be overridden by config
4) Tasks that aren't turned on as `Action Items` by default can still have an `ActionItemMessage`, in the event that they do become `Action Items`. That message is also overridable by config
5) Tasks that aren't `Action Items` by default can become `Action Items` by turning them on as such in the config. 
[...]

PR #488, opened 2020-06-11
Adding Actions to checkup

carakessler:
https://github.com/checkupjs/checkup/issues/475 for more info.

In this PR, I added the infrastructure to add action items to tasks, and render them in the output. There will be followup PRs adding availableActions
 to more tasks, and leveraging this feature further, but to start I added it to the eslint-disable task


packages/core/src/action-list.ts
```diff
@@ -0,0 +1,99 @@
+import { ActionConfig } from './types/config';
+import { Action } from './types/tasks';
+
+/**
+ * @class ActionList
+ *
+ * Represents a collection of actions related to a task
+ */
+export default class ActionList {
+  _availableActions: Action[];
```
scalvert, line 10: Should this just be a `Set` to make the filter/pop operations below easier?

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    return actionMessages;
+  }
+
+  get isActionable() {
+    if (this.actionMessages.length > 0) {
+      return true;
+    }
+
+    return false;
```
scalvert, line 54:
```suggestion
    return this.actionMessages.length > 0;
```

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,99 @@
+import { ActionConfig } from './types/config';
```
scalvert, line 1: There seems to be a lot of complexity in this class around available vs. enabled actions. It's a hard distinction to make from the reader's perspective.

Is there any reason we don't just take in the action config, immediately process it in the constructor, and then we only have the single list of actions, retrievable by key? If a task needs an action, it requests it by key. If there's a default for that action, it gets returned. If there's a custom configured action, it gets that. I'm sure I'm missing something here.

packages/core/src/base-task-result.ts
```diff
@@ -1,9 +1,8 @@
...
+import ActionList from './action-list';
 
 export default abstract class BaseTaskResult {
-  meta: TaskMetaData;
-
-  constructor(meta: TaskMetaData) {
-    this.meta = meta;
-  }
+  actionList!: ActionList;
+  constructor(public meta: TaskMetaData, public config: ActionConfig[]) {}
```
scalvert, line 7: I think it's fine if you want to pass actions config in, but if you do then it can't be assumed that `TaskConfig === ActionConfig`. I believe you're passing `task.config` directly into these tasks, assuming that the `TaskConfig` _is_ an `ActionConfig`. I think this is a mistake, as per my prior comment in this PR.

packages/core/src/base-task.ts
```diff
@@ -11,7 +11,7 @@ export default abstract class BaseTask {
   debug: debug.Debugger;
 
   #pluginName: string;
-  #config!: unknown;
+  #config!: ActionConfig[];
```
scalvert, line 14: Again, same issue with making `TaskConfig === ActionConfig`. They aren't the same.

packages/core/src/types/config.ts
```diff
@@ -1,4 +1,5 @@
-export type TaskConfig = 'on' | 'off' | ['on' | 'off', unknown];
+export type ActionConfig = string | { [key: string]: number } | { [key: string]: 'on' | 'off' };
+export type TaskConfig = 'on' | 'off' | ['on', ActionConfig[]];
```
scalvert, line 2: Same comment here re: `TaskConfig !== ActionConfig`.

packages/cli/src/reporters.ts
```diff
@@ -96,3 +88,57 @@ export function getReporter(
...
+          80,
+          { trim: false, hard: true }
+        ),
+        {
+          padding: 1,
+          borderStyle: 'double',
+        }
+      )
+    );
+    ui.blankLine();
```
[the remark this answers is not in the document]
carakessler: oooh good call ill move that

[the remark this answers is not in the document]
carakessler: and i can set options on `boxen` so all boxes will look the same (if we eventually have multiple)

packages/core/__tests__/action-list-test.ts
```diff
@@ -0,0 +1,95 @@
+import { Action, ActionConfig } from '../lib';
+import ActionList from '../lib/action-list';
```
scalvert, line 2: We shouldn't be referencing things from lib here. We should be referencing things from src, as per other tests.

packages/core/src/types/tasks.ts
```diff
@@ -21,17 +22,16 @@ export interface Task {
...
-export interface TaskContext {
-  readonly cliArguments: string[];
-  readonly cliFlags: RunFlags;
-  readonly parsers: Map<ParserName, CreateParser<ParserOptions, Parser<ParserReport>>>;
-  readonly config: CheckupConfig;
-  readonly pkg: PackageJson;
-  readonly paths: FilePathsArray;
+export interface Action {
+  name: string;
+  isEnabled: () => boolean;
```
scalvert, line 27: We should make this a getter.

```suggestion
  readonly isEnabled: boolean;
```

packages/core/src/types/tasks.ts
```diff
@@ -21,17 +22,16 @@ export interface Task {
...
+  threshold: number;
+  value: number;
+  message: () => string;
```
scalvert, line 30: We should make this a getter.

```suggestion
  readonly message: string;
```

packages/core/src/types/tasks.ts
```diff
@@ -21,17 +22,16 @@ export interface Task {
...
-  readonly cliFlags: RunFlags;
-  readonly parsers: Map<ParserName, CreateParser<ParserOptions, Parser<ParserReport>>>;
-  readonly config: CheckupConfig;
-  readonly pkg: PackageJson;
-  readonly paths: FilePathsArray;
+export interface Action {
+  name: string;
+  isEnabled: () => boolean;
+  threshold: number;
+  value: number;
```
scalvert, line 29: I'm not sure what the name `value` means here in terms of an `Action`. Can we have a better name so it's clear what we're trying to model here?

packages/core/__tests__/action-list-test.ts
carakessler: When I import from src I get the error 
```
Argument of type 'import("/Users/ckessler/checkup/packages/core/src/action-list").default' is not assignable to parameter of type 'import("/Users/ckessler/checkup/packages/core/lib/action-list").default'.
  Types have separate declarations of a private property 'findConfigForAction'.ts(2345)
```

Inspecting the files in lib, I dont see why this error is being thrown... thoughts?

carakessler: jk fixed

packages/core/src/types/tasks.ts
carakessler: Talked offline - going to leave this for now as we couldn't arrive at a better name 😨 

Going to make this clearer through documentation

packages/cli/__tests__/__utils__/mock-task-result.ts
```diff
@@ -1,8 +1,8 @@
-import { BaseTaskResult, TaskMetaData, TaskResult } from '@checkup/core';
+import { BaseTaskResult, TaskMetaData, TaskResult, ActionConfig } from '@checkup/core';
 
 export default class MockTaskResult extends BaseTaskResult implements TaskResult {
-  constructor(meta: TaskMetaData, public result: any) {
-    super(meta);
+  constructor(meta: TaskMetaData, config: ActionConfig[], public result: any) {
```
scalvert, line 4: `TaskResult` types take a `TaskConfig` in their constructor. This seems misaligned with that. Can we fix?

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -17,7 +17,7 @@ export class InsightsTaskHigh extends BaseTask implements Task {
   }
 
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'insights task high is being run');
+    return new MockTaskResult(this.meta, [], 'insights task high is being run');
```
scalvert, line 20: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -35,7 +35,7 @@ export class InsightsTaskLow extends BaseTask implements Task {
     super('fake', context);
   }
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'insights task low is being run');
+    return new MockTaskResult(this.meta, [], 'insights task low is being run');
```
scalvert, line 38: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -53,7 +53,7 @@ export class RecommendationsTaskHigh extends BaseTask implements Task {
     super('fake', context);
   }
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'recommendations task high is being run');
+    return new MockTaskResult(this.meta, [], 'recommendations task high is being run');
```
scalvert, line 56: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -71,7 +71,7 @@ export class RecommendationsTaskLow extends BaseTask implements Task {
     super('fake', context);
   }
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'recommendations task low is being run');
+    return new MockTaskResult(this.meta, [], 'recommendations task low is being run');
```
scalvert, line 74: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -89,7 +89,7 @@ export class MigrationTaskHigh extends BaseTask implements Task {
     super('fake', context);
   }
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'migration task high is being run');
+    return new MockTaskResult(this.meta, [], 'migration task high is being run');
```
scalvert, line 92: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -107,7 +107,7 @@ export class MigrationTaskLow extends BaseTask implements Task {
     super('fake', context);
   }
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'migration task low is being run');
+    return new MockTaskResult(this.meta, [], 'migration task low is being run');
```
scalvert, line 110: Pass task config in here.

packages/cli/__tests__/__utils__/mock-tasks.ts
```diff
@@ -144,6 +144,6 @@ export class TaskWithoutCategory extends BaseTask implements Task {
   }
 
   async run(): Promise<TaskResult> {
-    return new MockTaskResult(this.meta, 'task without category is being run');
+    return new MockTaskResult(this.meta, [], 'task without category is being run');
```
scalvert, line 147: Pass task config in here.

packages/cli/src/reporters.ts
```diff
@@ -96,3 +86,46 @@ export function getReporter(
       return async () => {};
   }
 }
+
+function getActionItems(pluginTaskResults: TaskResult[]): string[] {
+  return pluginTaskResults
+    .filter((taskResult) => taskResult.actionList?.isActionable === true)
```
scalvert, line 92:
```suggestion
    .filter((taskResult) => taskResult.actionList?.isActionable)
```

packages/cli/src/reporters.ts
```diff
@@ -96,3 +86,46 @@ export function getReporter(
...
+    .flatMap((actionableTask) => actionableTask.actionList?.actionMessages)
+    .filter(Boolean) as string[];
```
scalvert, line 94: I know we spoke about making this not just a string array. It's fine to leave it for now, but this will definitely change going forward.

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Represents a collection of actions related to a task
+ */
+export default class ActionList {
+  _availableActions: Action[];
+  _actionConfig: ActionConfig[] | [];
+
+  constructor(availableActions: Action[], config: TaskConfig | undefined) {
+    this._availableActions = availableActions;
+    if (Array.isArray(config)) {
+      this._actionConfig = [];
```
scalvert, line 16: I'd opt to remove this. I think you should _always_ pass an object with an `actions` property if you have actions. If one isn't present, we can default to empty array, as below.

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    return actionMessages;
+  }
+
+  get isActionable() {
+    return this.actionMessages.length > 0;
+  }
+
+  get enabledActions() {
```
scalvert, line 37: Can we just call this `actions`? It'd be implied that they're enabled, as we'd only provide once we'd determined to be enabled.

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        if (action.enabled) {
+          return action;
+        }
+      })
+      .filter(Boolean) as Action[];
+  }
+
+  private findConfigForAction(action: Action): ActionConfig | undefined {
```
scalvert, line 54: This is a private method, but for consumers of the JS transpiled version, there is no private designation, so we should use the conventional underscore.

```suggestion
  private _findConfigForAction(action: Action): ActionConfig | undefined {
```

packages/core/src/action-list.ts
```diff
@@ -0,0 +1,70 @@
...
+    return this._actionConfig?.find((conf) => Object.keys(conf)[0] === action.name);
+  }
+
+  private applyConfigForAction(action: Action, actionConfig: ActionConfig): Action | undefined {
```
scalvert, line 58: Same as above.

```suggestion
  private _applyConfigForAction(action: Action, actionConfig: ActionConfig): Action | undefined {
```

packages/core/src/types/config.ts
```diff
@@ -1,6 +1,7 @@
-export type TaskConfig = 'on' | 'off' | ['on' | 'off', unknown];
+export type ActionConfig = { [key: string]: number | 'off' };
```
scalvert, line 1: I think this should be changed to:

```suggestion
export type ActionConfig = { [key: string]: 'on' | 'off' | { threshold: number } };
```

I think this accounts for the following cases:
- For actions that are off by default (I know we discussed on by default, but we can't guarantee that all authors of tasks won't want to turn off tasks), you can turn them 'on'
- For actions that are on by default, you can turn them 'off'
- For actions that you want to configure the threshold for, you should explicitly set it, so that future readers of the config will know what that number represents without having to look it up in docs/read the code.

packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
```diff
@@ -86,4 +92,28 @@ describe('eslint-disable-task', () => {
...
   });
+
+  it('returns action item if there are more than 2 instances of eslint-disable', async () => {
+    const result = await new EslintDisableTask(
+      pluginName,
+      getTaskContext({
+        paths: project.filePaths,
+        config: {
+          tasks: {
+            'javascript/eslint-disables': ['on', ['numEslintDisables']],
```
scalvert, line 103: I think we discussed updating the `ActionConfig` so that it wasn't just a string to implicitly mark this as being 'on'.

I think we should have an here, with an actions property. That property can be an object of action configs, that we then read. 

```suggestion
            'javascript/eslint-disables': ['on', {
              'actions': {
                'numEslintDisables': 'on',
                'foo': {
                  'threshold': 10
                }
              }
            }],
```

packages/core/src/types/config.ts
```diff
@@ -1,6 +1,7 @@
...
+export type TaskConfig = { actions?: ActionConfig[]; [key: string]: any } | [];
```
scalvert, line 2: Based on the above change, I think `ActionConfig` should be an object of key (action name) and value ('on', 'off', or object that defines a property `threshold`, and a number value)

```suggestion
export type TaskConfig = { actions?: ActionConfig; [key: string]: any } | [];
```

packages/core/src/action-list.ts
carakessler: I disagree - there is input to the class we call "availableActions" which is of course private, but I think it is important to be clear on what `enabledActions` represents. It is NOT just the `actions` in the `ActionList`, it is the actions contained by the list that have currently evaluated to be enabled. I think the naming here is explicit and correct

packages/core/src/types/config.ts
carakessler: We actually can guarantee that authors of tasks wont turn off actions by default, because we removed the property where that could be set and that feature is not supported. The `ActionList` does not look for the word `on` and there isnt a property that can be set on an Action to turn it off. So there is no reason to account for `on` here, unless you are changing your mind and and want to allow tasks to be turned off by default (which I had originally in this PR as `enabledByDefault`). I can put that back in if that is what you are suggesting, but confused as to what else you could be referring to since we aligned that for this initial iteration, all tasks would be turned on by default

packages/core/src/action-list.ts
carakessler: compiler fails without this array check. i cleaned it up a bit and made it one line
 ```
    this._actionConfig = Array.isArray(config) ? [] : config?.actions || [];
```

if I do 
```
    this._actionConfig =  config?.actions || [];
```
I get 

```
Property 'actions' does not exist on type 'TaskConfig'. 
Property 'actions' does not exist on type '[]'.ts(2339)
```

File: packages/core/src/types/config.ts (hunk 28 above)
Comment (carakessler) [reply packages/core/src/types/config.ts:2]: Correct, but there can be multiple Actions configured per Task, so this still needs to be an array

File: packages/core/src/types/config.ts (hunk 26 above)
Comment (carakessler) [reply packages/core/src/types/config.ts:1]: Made the change to have threshold be an object instead of a raw number - agreed this is better for readability

File: packages/core/src/types/config.ts (hunk 28 above)
Comment (scalvert) [reply packages/core/src/types/config.ts:2]: I was thinking `ActionConfig` would be an object like so:

```
"actions": {
  "numFoos": { threshold: 5 },
  "bars": "off"
}
```

Why does it need to be an array when it can be an object keyed off the action name?

File: packages/core/src/action-list.ts (hunk 23 above)
Comment (scalvert) [reply packages/core/src/action-list.ts:33]: My earlier point about this class was that we should pass in the config, and only publicly expose what we need from it. You never get an action that's disabled from this class. Calling this `enabledActions` is superfluous; all actions returned here are implicitly enabled. You would never want, nor provide, any actions from the action list that were disabled. 

My assertion stands that we should change this.

File: packages/core/src/action-list.ts [hunk 29]
```diff
@@ -0,0 +1,67 @@
...
+ * @class ActionList
+ *
+ * Represents a collection of actions related to a task
+ */
+export default class ActionList {
+  _availableActions: Action[];
+  _actionConfig: ActionConfig;
+
+  constructor(availableActions: Action[], config: TaskConfig | undefined) {
+    this._availableActions = availableActions;
```
scalvert, line 14: I think this is more like `defaultActions`, isn't it? It's the defaults that you provide, which you then configure via the configuration, and provide as a list of actions.

packages/core/src/base-task.ts
```diff
@@ -44,9 +44,8 @@ export default abstract class BaseTask {
       return;
     }
 
-    let config: TaskConfig | undefined = this.context.config.tasks[this.fullyQualifiedTaskName];
+    let config = this.context.config.tasks[this.fullyQualifiedTaskName];
```
scalvert, line 47: I'm not sure this is correct. It seems we narrowed the scope of the `TaskConfig` type to not include the strings for "on" and "off", or the full representation of the tuple. 

What happens if there's no config? This will be undefined, and our config will end up undefined, possibly throwing and error if someone attempts to access it in a subclass.

packages/core/src/action-list.ts
carakessler: ya i can update that naming

packages/core/src/base-task.ts
carakessler:
```
    let config: 'on' | 'off' | ['on' | 'off', TaskConfig] | undefined = this.context.config.tasks[
      this.fullyQualifiedTaskName
    ];
```

carakessler:
```
    let config: 'on' | 'off' | ['on' | 'off', TaskConfig] | undefined = this.context.config.tasks[
      this.fullyQualifiedTaskName
    ];
```

this is correct here. I can add in those types to be very clear - its the only place it needs to be used, so I think its ok to leave it like this. We could add a new type for `'on' | 'off' | ['on' | 'off', TaskConfig]` but seems unnecessary since its only used here, and hard to name lol. I think its important that the TaskConfig definition stay as is, because that is what the tasks pass around. The `on/off` piece of the logic is encapsulated here in `base-task` and authors don't need to know anything about it

scalvert: The reason I named it TaskConfig - which represents all of the config, including the enabled setting, is because the entire portion of this is the config. It's the value of the key that is the task name.

I appreciate that your take on it is that the actual configuration, the previously `unknown` portion, is the config itself. I'm fine to keep it as it is in this PR. If it makes less sense to me later, we can always change it back.

scalvert:
Just going to merge this for now. We can iterate on tidying anything that's rough later.

PR #498, opened 2020-06-15
Removing TaskType in favor of group

scalvert:
This removes the `TaskType` enum and `type` prop in `TaskClassification` in favor of a `group` string prop. This provides more flexibility to specify both the category (primary grouping mechanism) and group (secondary grouping mechanism).

Example: under the `linting` category, you could specify an `eslint` group comprised of 3 tasks and a `template lint` group comprised of 2 tasks. The sorting would be as follows:

```
linting

eslint task 1
eslint task 2
eslint task 3

template lint task 1
template lint task 2 
```

TODO

- [ ] Add sorting secondarily by group (after category)


packages/cli/src/generators/task.ts
```diff
@@ -82,7 +68,6 @@ export default class TaskGenerator extends BaseGenerator {
     this.options.taskClass = `${this.options.pascalCaseName}Task`;
     this.options.taskResultClass = `${this.options.taskClass}Result`;
     this.options.typescript = this.answers.typescript;
-    this.options.type = this.answers.type;
```
carakessler, line 85: why isn't group a part of the generator?

packages/cli/src/generators/task.ts
```diff
@@ -59,17 +56,6 @@ export default class TaskGenerator extends BaseGenerator {
...
-          choices: [
-            { name: 'insights', value: 'Insights' },
-            { name: 'migrations', value: 'Migrations' },
-            { name: 'recommendations', value: 'Recommendations' },
-          ],
-          default: 'Insights',
-        },
         {
           type: 'input',
           name: 'category',
```
carakessler, line 61: since category is essentially acting as `type` and is now the top level sorting mechanism, i think we need more description and guidance here than "Enter a task category"... maybe show some examples of categories?

packages/cli/templates/src/task/src/tasks/task.js.ejs
```diff
@@ -1,12 +1,11 @@
-import { BaseTask, TaskType } from '@checkup/core';
+import { BaseTask } from '@checkup/core';
 import <%- taskResultClass %> from '../results/<%- name %>-task-result';
 
 export default class <%- taskClass %> extends BaseTask {
   meta = {
     taskName: '<%- name %>',
     friendlyTaskName: '<%- _.startCase(name.split('-').join(' ')) %>',
     taskClassification: {
-      type: TaskType.<%- _.upperFirst(type) %>,
```
carakessler, line 9: don't we want group here?

packages/cli/templates/src/task/src/tasks/task.ts.ejs
```diff
@@ -1,12 +1,11 @@
-import { BaseTask, TaskType, Task, TaskMetaData, TaskResult } from '@checkup/core';
+import { BaseTask, Task, TaskMetaData, TaskResult } from '@checkup/core';
 import <%- taskResultClass %> from '../results/<%- name %>-task-result';
 
 export default class <%- taskClass %> extends BaseTask implements Task {
   meta: TaskMetaData = {
     taskName: '<%- name %>',
     friendlyTaskName: '<%- _.startCase(name.split('-').join(' ')) %>',
     taskClassification: {
-      type: TaskType.<%- _.upperFirst(type) %>,
```
carakessler, line 9: and here

packages/cli/templates/src/task/src/tasks/task.js.ejs
carakessler: i know its optional and people can remove it if they want, but i feel like it should be a part of the output from the generators

packages/cli/src/generators/task.ts
scalvert: Because it's optional.

packages/cli/src/generators/task.ts
scalvert: Well the idea is that they're free text. I hesitate to provide a static list of categories, as we want people to have flexibility in what they use here.

packages/cli/templates/src/task/src/tasks/task.js.ejs
scalvert: I disagree. Generators should be generating the bare minimum. You can add the group later, if needed.

packages/cli/templates/src/task/src/tasks/task.ts.ejs
scalvert: Same same.

packages/cli/src/generators/task.ts
carakessler: Even if it is optional, cant we provide a step for authors to enter it, indicating that it is optional and can be skipped?

packages/cli/src/generators/task.ts
carakessler: maybe a bit more info about what the category is used for perhaps? "Categories are used to group similar tasks together to help organize the results. Some examples of categories are: 'Best Practices', 'Testing', etc."

packages/cli/src/generators/task.ts
scalvert: Ya I thought twice about this, and think it does make sense as a discoverability mechanism. I've added group.

packages/cli/templates/src/task/src/tasks/task.js.ejs
scalvert: I had second thoughts about this, as per my previous comment. Added now.

packages/cli/src/generators/task.ts
scalvert: Added!

PR #499, opened 2020-06-16
Adding actions for ember-test-types, template-lint, and outdated dependencies

carakessler:


packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -42,33 +51,87 @@ export default class EmberTestTypesTaskResult extends BaseTaskResult implements
...
+          get enabled() {
+            return this.value > this.threshold;
+          },
+          get message() {
+            return `${fractionToPercent(
+              totalSkips,
+              totalTests
+            )} of your tests are skipped, this value should be below ${decimalToPercent(
+              this.threshold
+            )}`;
```
scalvert, line 115: You and I have a bit of a different idea about what this should look like. I believe actions should be comprised of two main things:

1. **The action** - written in the declarative voice: "Reduce number of skipped tests". This should be a short, to-the-point message.
1. **Additional information** - written with more detail: "Total skipped tests is 10%. The current threshold is 2%"
1. **Detail** - the actual list of tests that are skipped. This obviously doesn't work well for stdout, but is targeted at the report viewer (think hidden, scrollable div with a table).

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -42,33 +51,87 @@ export default class EmberTestTypesTaskResult extends BaseTaskResult implements
...
+      }
+      return total;
+    }, 0);
+    let totalApplication =
+      this.testTypes.filter((testType) => testType.type === TestType.Application).pop()?.total || 0;
+
+    return new ActionList(
+      [
+        {
+          name: 'percentageSkippedTests',
```
scalvert, line 103: We should think about whether we want a structure for these action names. For instance, do we want to prefix with the task, followed by a name?

```
ember-test-types:percentageOfSkippedTests
```

carakessler: so in the config, that would read pretty weirdly, as it would look like: 
```
tasks: 
    'ember:ember-test-types': { actions: { 'ember-test-types:percentageOfSkippedTests': 'off' } },
  };
```

in the code path coming in from the config - we know what the actions are meant to be associated with. The Actions are also contained in an ActionList that is a property of the task itself, so again, we know what task the action is associated with. I don't think any sort of additional structure is necessary here, but lmk what you think!

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
carakessler: Are you proposing that we break the message up into an object of those 3 types instead of it being a string? 

I agree that there will definitely need to be an overview of the tone and thresholds of all the actions (as we discussed offline), just wanted to get them all in and see what kinds of actions we have, what sort of data is being collected, before that level of polish.

scalvert: Yes. I've mentioned this before, that a simple `string[]` isn't sufficient to convey the info that we should be presenting. I think the action result should be a more complex object that essentially is comprised of the above fields:

```
interface ActionResult {
  message: string;
  violation: string;
  details: any;
}
```

Let's pause this for now, but I do want to circle back on this in a subsequent PR to cleanup the output.

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
scalvert: Makes sense. What I want to avoid is adhoc, hard to understand names that don't do a good job indicating to the user what they do. 

At the least, I think we should keep them hyphen case, so we don't mix naming types (camel and hyphen). Would you be OK with making that change?

carakessler: ya sure i'll update all action names now

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -42,33 +51,87 @@ export default class EmberTestTypesTaskResult extends BaseTaskResult implements
...
+    let totalSkips = this.totalSkips;
+
+    let totalRenderingUnit = this.testTypes.reduce((total, item: TestTypeInfo) => {
+      if (item.type === TestType.Rendering || item.type === TestType.Unit) {
+        return total + item.total;
+      }
+      return total;
+    }, 0);
+    let totalApplication =
+      this.testTypes.filter((testType) => testType.type === TestType.Application).pop()?.total || 0;
```
scalvert, line 98:
```suggestion
      this.testTypes.find((testType) => testType.type === TestType.Application).total || 0;
```

packages/checkup-plugin-ember/src/results/ember-test-types-task-result.ts
```diff
@@ -42,33 +51,87 @@ export default class EmberTestTypesTaskResult extends BaseTaskResult implements
...
-  return tableColumns;
+  get actionList() {
+    let totalTests = getTotalOfType(this.testTypes, 'total');
+    let totalSkips = this.totalSkips;
+
+    let totalRenderingUnit = this.testTypes.reduce((total, item: TestTypeInfo) => {
+      if (item.type === TestType.Rendering || item.type === TestType.Unit) {
+        return total + item.total;
+      }
+      return total;
```
scalvert, line 95:
```suggestion
      return (item.type === TestType.Rendering || item.type === TestType.Unit) ? total + item.total: total;
```

PR #500, opened 2020-06-16
Removing 'on' option from task config because tasks are always on by default

carakessler:
As tasks are always 'on' by default, there is no reason to have a config option to turn them on. If that changes, we will add this option back in. This change aligns with how the ActionConfig works

### DIFF
--- a/packages/core/src/base-task.ts
+++ b/packages/core/src/base-task.ts
@@ -44,7 +44,7 @@ export default abstract class BaseTask {
       return;
     }
 
-    let config: 'on' | 'off' | ['on' | 'off', TaskConfig] | undefined = this.context.config.tasks[
+    let config: 'off' | ['off', TaskConfig] | TaskConfig | undefined = this.context.config.tasks[
       this.fullyQualifiedTaskName
     ];
 
@@ -57,6 +57,8 @@ export default abstract class BaseTask {
 
       this.#enabled = enabled;
       this.#config = taskConfig;
+    } else {
+      this.#config = config;
     }
 
     this.debug('%s enabled: %s', this.constructor.name, this.#enabled);
--- a/packages/core/__tests__/base-task-test.ts
+++ b/packages/core/__tests__/base-task-test.ts
@@ -45,7 +45,7 @@ describe('BaseTask', () => {
       config: {
         plugins: [],
         tasks: {
-          'fake/my-fake': ['on', [{ 'my-fake-option': 20 }]],
+          'fake/my-fake': { 'my-fake-option': 20 },
         },
       },
     });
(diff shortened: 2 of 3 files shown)


packages/core/src/base-task.ts
```diff
@@ -44,7 +44,7 @@ export default abstract class BaseTask {
       return;
     }
 
-    let config: 'on' | 'off' | ['on' | 'off', TaskConfig] | undefined = this.context.config.tasks[
+    let config: 'off' | ['off', TaskConfig] | TaskConfig | undefined = this.context.config.tasks[
```
scalvert, line 47: I think we should define a type alias for this. Can we add one? Though unfortunately, I think `TaskConfig` is still the best name for this...

You could make it `TaskConfigOrEnabled`, which I've seen in some type names before when you're trying to represent intersections.

carakessler: Are you suggesting I create: 
```
type TaskConfigOrEnabled = 'off' | ['off', TaskConfig] | TaskConfig;
```

That naming is confusing to me, it makes me think it would be: 
```
type TaskConfigOrEnabled = 'off' | TaskConfig;
```

I agree that there would be value in creating a type here, i just couldnt think of a useful/meaningful name....

carakessler: I was thinking we could do something like I did for `ActionConfig` 
```
export type ActionConfigValue = { threshold: number } | 'off' | ['off', { threshold: number }];
export type ActionConfig = Record<string, ActionConfigValue>;
export type TaskConfigValue = { actions?: ActionConfig; [key: string]: any };
export type TaskConfig = 'off' | ['off', TaskConfigValue] | TaskConfigValue;
export type CheckupConfig = {
  excludePaths: string[];
  plugins: string[];
  tasks: Record<string, TaskConfig>;
};
```

To me, this makes the most sense, but then all over the repo we will be using `TaskConfigValue` which is the shittier name, to then get to use `TaskConfig` only once, which is why I ended up not creating a type here and just throwing the whole union in. Worth noting I also could just remove the type info period on line 47, as the editor already knows what it will be... 

it would just read 
```
    let config = this.context.config.tasks[this.fullyQualifiedTaskName];
``` 
and then when hovering over `config` I can see correct type info. Then we would clean up the stuff here, and not need to create an additional type. Let me know what you think is preferable

scalvert: It's true that IDEs help here, but not everyone is using an IDE (vim/emacs), and this doesn't help us when we're reading code reviews. I do, however, think it's fine to just leave it as is since, as you say, it's only used in one place.

PR #503, opened 2020-06-17
Sorting tasks by category, then group.

scalvert:
Dependent Tasks:

- [ ] Add default categories to the docs site.

### DIFF
--- a/packages/cli/src/task-result-comparator.ts
+++ b/packages/cli/src/task-result-comparator.ts
@@ -1,0 +1,35 @@
+import { TaskClassification, TaskResult } from '@checkup/core';
+
+const DEFAULT_CATEGORIES: Record<string, number> = {
+  metrics: 6,
+  'best practices': 5,
+  dependencies: 4,
+  linting: 3,
+  testing: 2,
+  migrations: 1,
+};
+
+function getCategorySort(category: string): number {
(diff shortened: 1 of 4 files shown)


packages/cli/src/task-result-comparator.ts
```diff
@@ -0,0 +1,35 @@
+import { TaskClassification, TaskResult } from '@checkup/core';
+
+const DEFAULT_CATEGORIES: Record<string, number> = {
+  metrics: 6,
```
carakessler, line 4: I think we would want this prioritization to be configurable, no?  

also - if we are going to "rate"  what we consider to be the "default categories", we definitely need to document this somewhere so people are _somewhat_ encouraged to adhere to default categories

carakessler:
Can we add tests for this por favor?

scalvert:
Already on it! I wouldn't leave home without tests :)

packages/cli/src/task-result-comparator.ts
scalvert: I don't think we want default categories configurable. I do believe we should document the default categories. That can be added to the docs site.

scalvert:
Added tests @carakessler. I think this is GTG - I will add documentation after this lands.

PR #515, opened 2020-06-23
Ensuring generators can generate into a specific path

scalvert:
Fixes generators to correctly generate in a specified path. Changes the path arg to a flag.

```bash
checkup generate config --path ./lib
```

TODO:

- [x] Add tests

### DIFF
--- a/packages/cli/src/commands/generate.ts
+++ b/packages/cli/src/commands/generate.ts
@@ -44,7 +44,6 @@ export default class GenerateCommand extends BaseCommand {
     {
       name: 'name',
       description: 'name of the entity (kebab-case)',
-      optional: true,
     },
   ];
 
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/generate.ts
```diff
@@ -39,11 +44,7 @@ export default class GenerateCommand extends BaseCommand {
     {
       name: 'name',
```
carakessler, line 45: why is `name` now optional?

carakessler: it seems like it would be optional maybe for generating a `config` but required for task and plugin? how would this play out for user?

scalvert: Yes, it was optional due to config. We'll never document it being optional. It wouldn't ever make sense for anyone to run the `plugin` or `task` generators without a name.

carakessler: will it throw an error if they do run plugin || task without a name? i feel like it should.. just want to make sure this is as straightforward as possible to users

scalvert: I've actually reversed this, on second look. The config generator works fine without name being optional, and it helps enforce the other two generators functionality. Thanks for pointing this out.

PR #520, opened 2020-06-26
Refactors tasks to have correct isolation between Task and TaskResult classes

scalvert:
The current implementation of `Task` and `TaskResult` is a bit loose, in that we currently implement code in `Task.run` that has intimate details of the `TaskResult` class' state. Specifically, we set arbitrary properties on the `TaskResult` class via the `Task`. This is a class boundary violation, in that the `Task` class author shouldn't need to know specifically the details of the respective properties in the associated `TaskResult`.

This PR addresses this in the following ways:
1. Adds a `TaskResult.process` method, which accepts an arbitrary data bag
1. All result processing, including property setting, transformations, and actions instantiation, should occur in the `process` method

New `TaskResult` interface:

```ts
export interface TaskResult {
  meta: TaskMetaData;
  data: Record<string, any>;
  actionList?: ActionList;

  process(data: Record<string, any>): void;
  toConsole: () => void;
  toJson: () => JsonMetaTaskResult | JsonTaskResult;
}
```

This allows `Task` classes to be responsible for one thing: gathering data. This also allows `TaskResult` classes to be responsible for two things: transforming and displaying data.

Updates all tasks and task result classes and associated tests to adhere to this new pattern.

cc: As discussed with @rwjblue 

Note: there will likely be subsequent refactorings that further clean up these classes.


scalvert:
I'm going to go ahead and merge this to keep the refactoring flowing.

PR #524, opened 2020-06-29
[tests] add tests and fixes sectionedBar calculations

gabrielcsapo:
*Motivation*

I was using sectionedBar in my task and it was resulting in funky output. 

# Before

Screen Shot 2020-06-30 at 11 53 28 AM

# After

Screen Shot 2020-06-30 at 11 59 16 AM


scalvert:
Thanks, @gabrielcsapo!

scalvert:
It looks like we have some failing tests. Could you take a look?

PR #526, opened 2020-06-30
Adds a specific type for a config value

scalvert:
This PR adds a specific type to represent the common config value pattern we use:

```typescript
type ConfigValue<T extends {}> = 'on' | 'off' | ['on' | 'off', T];
```

This allows us to better represent this common pattern when used in code.

Additionally, I've added a method to handle the common parsing of this type of value:

```typescript
function getConfigTuple<T>(configValue: ConfigValue<T> | undefined): [boolean, T]
```

This allows us to parse the config tuple in a consistent manner, without repeating our code everywhere.

**Note**: This reverses changes to some of the types we made recently, namely the removal of the `'on'` value in config. This is mainly to allow for us to release new tasks/actions etc without having them explicitly enabled. A future PR will introduce the concept of default configuration for tasks, which will allow us to specify whether a task should be enabled by default or not.

### DIFF
--- a/packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
+++ b/packages/checkup-plugin-ember/__tests__/ember-test-type-task-test.ts
@@ -184,7 +184,19 @@ describe('ember-test-types-task', () => {
         paths: project.filePaths,
         config: {
           tasks: {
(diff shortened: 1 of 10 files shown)


scalvert:
cc @rwjblue 

As discussed in our pairing session, I'm adding this change as a path to our refactor of the actions functionality. I wanted to land this smaller increment so as to keep that other PR small.

PR #569, opened 2020-07-22
Converting to new checkup schema format.

scalvert:
Based on https://github.com/checkupjs/rfcs/pull/1

This is the first set of changes of a number of successive changes that will move us toward the standardized JSON schema defined in RFC-1. I want to keep these changes small and incremental, so they're easier to review.

### DIFF
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -1,4 +1,5 @@
 import { BaseTask, Task, TaskResult, SummaryData } from '@checkup/core';
+import { formatSummary } from '@checkup/core';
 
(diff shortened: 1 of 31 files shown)


packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
```diff
@@ -70,18 +88,32 @@ function findDependencies(
...
+    .filter(Boolean) as Dependency[];
+}
+
+function formatSummaryData(key: string, data: Array<string | object>): SummaryData {
+  return {
+    key,
+    type: 'summary',
+    data,
+    count: Array.isArray(data) ? data.length : Object.keys(data).length,
+  };
```
scalvert, line 116: Ultimately this function should be pulled out to a centralized util. All schema formats should have associated utils that allow you to format the data to adhere to the schema. This will be done in a later PR.

Issue #525, opened 2020-06-29
[question] is there a way to run multiple tasks?


gabrielcsapo:
*Motivation*

I am looking to run a select group of tasks. 

When I run `checkup --task foo --task bar --task moo` this does not result in multiple tasks, just only moo. Would there be a way to run `checkup --tasks 'foo, bar, moo'` ?

scalvert:
Thanks, @gabrielcsapo. Currently there isn't a way to run multiple. We can pretty easily extend this to allow multiples to be run though. I'll add it to the backlog.

carakessler:
@gabrielcsapo I will be doing this today/tomorrow, will hit you up when it is done!

PR #609, opened 2020-08-06
Adding functionality to run multiple tasks via command line

carakessler:
https://github.com/checkupjs/checkup/issues/525


packages/cli/src/task-list.ts
```diff
@@ -71,6 +71,36 @@ export default class TaskList {
     return this.getTasks().find((task) => task.meta.taskName === taskName);
   }
 
+  /**
+   * Finds tasks by task name
+   *
+   * @method findTasks
+   * @param taskNames The name of the task to find
+   */
+  findTasks(taskNames: TaskName[]): { tasksFound: Task[]; tasksNotFound: TaskName[] } {
```
scalvert, line 80: Could we change this to use varargs instead? 

```suggestion
  findTasks(...taskNames: TaskName[]): { tasksFound: Task[]; tasksNotFound: TaskName[] } {
```

packages/cli/src/commands/run.ts
```diff
@@ -126,26 +137,21 @@ export default class RunCommand extends BaseCommand {
   private async runTasks() {
     [this.metaTaskResults, this.metaTaskErrors] = await this.defaultTasks.runTasks();
 
-    if (this.runFlags.task !== undefined) {
-      let pluginTaskResult: TaskResult | undefined;
-      let taskFound: boolean = false;
+    if (this.runFlags.tasks !== undefined) {
+      let { tasksFound, tasksNotFound } = this.pluginTasks.findTasks(this.runFlags.tasks);
```
scalvert, line 141: If we change the signature of `findTasks` as per my suggestion below, I think you may need to spread this?

```suggestion
      let { tasksFound, tasksNotFound } = this.pluginTasks.findTasks(...this.runFlags.tasks);
```

packages/cli/src/reporters/console-reporter.ts
```diff
@@ -180,6 +180,14 @@ let outputMap: { [taskName: string]: (taskResult: TaskResult) => void } = {
       );
     });
   },
+  foo: function (taskResult: TaskResult) {
```
scalvert, line 183: Such foo...

carakessler: lollllll

PR #635, opened 2020-08-13
Getting paths in CLI via threads

scalvert:
The UI is frozen when running the CLI, which gives the impression that things are stalled. This defeats the entire purpose of the UI spinner, which is supposed to indicate that work is in progress.

Adding a worker thread for the walkSync where we expand globs helps with this.


scalvert:
cc/ @gabrielcsapo

PR #645, opened 2020-08-17
Prioritizing cli args for turning tasks on

carakessler:
Currently, if a task `foo` is set to "off" in config, but checkup is run via `checkup --task foo` the task doesnt run. The command line args override config for excludepaths, and it should work the same way for tasks. Now, if a task is passed in via command line arg, it will always run, regardless of config value

### DIFF
--- a/packages/core/src/base-task.ts
+++ b/packages/core/src/base-task.ts
@@ -16,7 +16,7 @@ export default abstract class BaseTask {
 
   _pluginName: string;
   _config!: TaskConfig;
-  _enabled!: boolean;
+  _enabledViaConfig!: boolean;
 
   constructor(pluginName: string, context: TaskContext) {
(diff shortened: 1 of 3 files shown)


packages/core/src/base-task.ts
```diff
@@ -34,7 +34,11 @@ export default abstract class BaseTask {
   get enabled() {
     this._parseConfig();
 
-    return this._enabled;
+    let enabledViaCli = this.context.cliFlags.task?.includes(this.fullyQualifiedTaskName) || false;
```
scalvert, line 37: I believe this can be enabled via the node API too. Can we call this something else to better capture that?

```suggestion
    let enabledViaFlag = this.context.cliFlags.task?.includes(this.fullyQualifiedTaskName) || false;
```

PR #649, opened 2020-08-18
Bug fix for ember-test-types-task

carakessler:
Bug in ember-test-types is triggered when a test file has `// eslint-disable-next-line` with a rule that our eslint parser doesnt have. Previously, it assumed that the message is going to contain results from the test-type rule but when eslint encounters a disable like that it  will throw an error w message
```
message: "Definition for rule '@linkedin/pemberly/disallow-classnames-in-assertions' was not found."
```
filtered lint results to ensure we are only looking at the data we want here


scalvert:
Thank you!

PR #651, opened 2020-08-18
Fixing eslint-disable-task to use recast (with correct babel plugins) vs. raw babel

scalvert:
Using babel parse/traverse requires us to configure the babel plugins manually. Recast already does this automatically, so switching back to using that.

### DIFF
--- a/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
+++ b/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
@@ -50,6 +50,13 @@ describe('eslint-disable-task', () => {
             "count": 3,
             "data": Array [
               Object {
+                "column": 19,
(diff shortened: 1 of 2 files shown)


packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
```diff
@@ -50,23 +50,23 @@ describe('eslint-disable-task', () => {
             "count": 3,
             "data": Array [
               Object {
-                "column": 4,
```
carakessler, line 53: @scalvert do you have any idea as to why these numbers are changing with this change?

scalvert: Yes, the object is serialized different, but the values are the same if you compare them.

PR #653, opened 2020-08-20
Adding eslint config for no-extraneous-imports. Fixes all missing imports

scalvert:
Adds eslint config to detect missing imports. Also adds all missing imports to specific packages.


.eslintrc
```diff
@@ -26,13 +39,51 @@
...
+        ]
+      }
+    },
+    {
+      "files": ["packages/checkup-plugin-javascript/src/**/*.ts"],
+      "rules": {
+        "node/no-extraneous-import": [
+          "error",
+          {
+            "allowModules": ["@babel/types", "ast-types", "type-fest"]
```
carakessler, line 73: why do we need this?

scalvert: Our type based modules are in the root package.json (as they should be). If we don't explicitly allow them, they will error. We only need to express these dependencies if we're using them in published interfaces, hence I'm just marking them as allowed. Does this make sense?

PR #751, opened 2020-11-11
Bug fixes

carakessler:
1. Fix runCommand so it captures all of stdout, rather than just last line
2. Export function renderEmptyResult so other reporters can use it
3. TaskReporter takes in array of Results, not just one Result
4. Adding additional helper used a lot in an upgrade of another plugin

### DIFF
--- a/packages/core/src/data/builders.ts
+++ b/packages/core/src/data/builders.ts
@@ -1,6 +1,6 @@
-import { ESLintMessage, ESLintReport } from '../types/parsers';
+import { ESLintMessage, ESLintReport, ESLintResult } from '../types/parsers';
 import { LintResult, TaskError } from '../types/tasks';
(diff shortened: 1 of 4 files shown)


packages/core/src/data/builders.ts
```diff
@@ -149,6 +149,21 @@ export function buildLintResultDataItem(
   };
 }
 
+export function buildLintResultsFromEslintOrTemplateLint(
```
scalvert, line 152: A bit wordy of a name?

scalvert (approved):

packages/core/src/data/builders.ts
carakessler: we can just have it be `buildLintResults` (because the params are obvious) but felt like that wasnt descriptive enough

PR #770, opened 2020-11-16
Reverting version of eslint-plugin-ember

scalvert:
The `eslint-plugin-ember` package in `checkup-plugin-ember-octane` is a dependency, as it relies on rules within that package to perform the data gathering for the octane task. As such, upgrades to this package can result in the absolute numbers that the task gathers changing based on changes to the plugin itself. 

In a correct versioning world (post 1.0.0), we'd signal these upgrades using a major version bump. Since we're in pre-release, we can't rely on that signal right now. For now, we'll just have to scrutinize changes to ensure we don't bump when we don't intend to.

To help with this, I've added a dependabot config to ignore updates to the package. Any bumps will need to be intentional going forward.


carakessler (approved):

.dependabot/config.yml
```diff
@@ -0,0 +1,12 @@
...
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "daily"
+    ignored_updates:
+      - match:
+          dependency_name: "eslint-plugin-ember"
+          # ignores any updates 9.0.0 or greater
+          # since we want to explicitly upgrade this
+          # dependency ourselves (it changes the nature of data tasks)
+          version_requirement: "^9.0.0"
```
chriskrycho, line 12: 💙

chriskrycho (approved):
Thank you!

.dependabot/config.yml
scalvert: This is going to be a temporary fix, as we're likely going to move this dependency to a peerDependency, which will mean checkup will use whatever the host app uses. This means this dependabot config will no longer be needed.

PR #777, opened 2020-11-18
Implements new node API to allow for programmatic invocation of Checkup

scalvert:
The stop-gap API implemented in (#705) is superseded by this PR. This uses the standard APIs that oclif provides, and allows for correct invocation of Checkup via node.

TODO

- [ ] Add tests


carakessler (approved):
thank you for getting this did 👊

PR #781, opened 2020-11-23
Fixing SARIF log so that each location is mapped to one Result

carakessler:
Even though locations is technically a Location[], if you read the SARIF spec it says that The locations array SHALL NOT be used to specify distinct occurrences of the same result which can be corrected independently.

Modifying tasks to return one Result per location.

Also renamed `toJson` to `appendCheckupProperties` because the name `toJson` made no sense :) In the future, I would love if we could remove this extra call all together, but that may have to involve putting the builders in the base class (or passing in all 4 properties we need to all the builders) but wanted to do that as a follow up to simplify this 

https://docs.oasis-open.org/sarif/sarif/v2.0/csprd02/sarif-v2.0-csprd02.html#_Toc10127841

### DIFF
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -1,5 +1,5 @@
 import { BaseTask, Task } from '@checkup/core';
-import { buildResultFromProperties } from '@checkup/core';
+import { buildResultsFromProperties } from '@checkup/core';
 
 import { PackageJson } from 'type-fest';
 import { Result } from 'sarif';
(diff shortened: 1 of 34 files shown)


packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
```diff
@@ -18,37 +18,37 @@ export default class EmberDependenciesTask extends BaseTask implements Task {
...
+    let coreLibraries = buildResultsFromProperties(
       [
         findDependency(packageJson, 'ember-source'),
         findDependency(packageJson, 'ember-cli'),
         findDependency(packageJson, 'ember-data'),
       ],
       'ember core libraries'
-    );
-    let emberDependencies = buildResultFromProperties(
+    ).map((result) => this.appendCheckupProperties(result));
```
scalvert, line 28: We seem to do this pattern a lot, where we `buildResultsFromProperties` and then we `map` `result` over that. Should we consider combining these, like have `buildResultsFromProperties` take `result` and `map` internally?

scalvert (changes requested):

scalvert (approved):

packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
scalvert: I'm fine if you want to do this in a follow up.

carakessler: i feel like the better way to fix this is:  

```
Also renamed toJson to appendCheckupProperties because the name toJson made no sense :) 
In the future, I would love if we could remove this extra call all together,
but that may have to involve putting the builders in the base class (or passing in all 4 
properties we need to all the builders) but wanted to do that as a follow up to simplify this
```

thoughts?

scalvert: Ya either works. Reducing repetitive code would really be the goal here.

PR #802, opened 2020-12-03
Add ability to run tasks by category/group

scalvert:
This change expands the existing functionality of the command line options to include the ability to run tasks by either category or group.

Similar to the `task` option, you can now pass a `category` option (multiple times):

```shell
checkup --category 'best practices'
```

or a `group` option (multiple times)

```shell
checkup --group 'ember'
```

### DIFF
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -79,17 +79,17 @@ export default class RunCommand extends BaseCommand {
       description: 'The path referring to the root directory that Checkup will run in',
     }),
     category: flags.string({
-      description: 'Runs specific task specified by category. Can be used multiple times.',
+      description: 'Runs specific tasks specified by category. Can be used multiple times.',
       multiple: true,
     }),
     group: flags.string({
-      description: 'Runs specific task specified by group. Can be used multiple times.',
+      description: 'Runs specific tasks specified by group. Can be used multiple times.',
       multiple: true,
     }),
     task: flags.string({
       char: 't',
       description:
-        'Runs specific task specified by the fully qualified task name in the format pluginName/taskName. Can be used multiple times.',
+        'Runs specific tasks specified by the fully qualified task name in the format pluginName/taskName. Can be used multiple times.',
       multiple: true,
     }),
     format: flags.string({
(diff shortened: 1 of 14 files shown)


packages/cli/src/commands/run.ts
```diff
@@ -77,6 +78,14 @@ export default class RunCommand extends BaseCommand {
       char: 'd',
       description: 'The path referring to the root directory that Checkup will run in',
     }),
+    category: flags.string({
+      description: 'Runs specific task specified by category. Can be used multiple times.',
```
carakessler, line 82: nit: should be `tasks`

packages/cli/src/commands/run.ts
```diff
@@ -77,6 +78,14 @@ export default class RunCommand extends BaseCommand {
...
+      multiple: true,
+    }),
+    group: flags.string({
+      description: 'Runs specific task specified by group. Can be used multiple times.',
```
carakessler, line 86: samezies

packages/cli/src/commands/run.ts
```diff
@@ -114,6 +123,18 @@ export default class RunCommand extends BaseCommand {
   checkupConfig!: CheckupConfig;
   cliModeEnabled: boolean = true;
 
+  get taskFilterType() {
+    if (this.runFlags.task !== undefined) {
+      return 'task';
+    } else if (this.runFlags.category !== undefined) {
```
carakessler, line 129: we should specify that groups and category flags cant be used together if this is going to be the business logic (FWIW i think its fine to have it be one or the other, would just want to make that clear in docs)

carakessler (approved):
Overall LGTM, just a few nits on documentation

packages/cli/src/commands/run.ts
scalvert: Fixed!

packages/cli/src/commands/run.ts
scalvert: Fixed!

packages/cli/src/commands/run.ts
scalvert: This is a good call. I'll make sure it's called out in the docs when I update them.

PR #812, opened 2020-12-07
Creating interface "TaskMetadata" and bucketing info about each task under it

carakessler:
This is PR 1 of 2 to remove the need for the `appendCheckupProperties` function pattern. After bucketing `taskName`, `taskDisplayName`, `category`, and `group` under `taskMetadata` I will put up another PR where I pass that data into the builders directly, and remove the need for the `appendCheckupProperties` on each Result

### DIFF
--- a/packages/checkup-plugin-ember-octane/src/tasks/ember-octane-migration-status-task.ts
+++ b/packages/checkup-plugin-ember-octane/src/tasks/ember-octane-migration-status-task.ts
@@ -79,10 +79,12 @@ const OWN_PROPERTIES_RULES = ['no-implicit-this'];
 const USE_MODIFIERS_RULES = ['no-action'];
 
 export default class EmberOctaneMigrationStatusTask extends BaseTask implements Task {
-  taskName = 'ember-octane-migration-status';
-  taskDisplayName = 'Ember Octane Migration Status';
-  category = 'migrations';
-  group = 'ember';
+  taskMetadata = {
+    taskName: 'ember-octane-migration-status',
+    taskDisplayName: 'Ember Octane Migration Status',
+    category: 'migrations',
+    group: 'ember',
+  };
 
   private eslintParser: Parser<ESLintReport>;
   private templateLinter: TemplateLinter;
--- a/packages/checkup-plugin-ember/__tests__/__snapshots__/ember-template-lint-summary-task-test.ts.snap
+++ b/packages/checkup-plugin-ember/__tests__/__snapshots__/ember-template-lint-summary-task-test.ts.snap
@@ -22,7 +22,7 @@ Array [
     "occurrenceCount": 1,
     "properties": Object {
       "category": "linting",
-      "group": "ember",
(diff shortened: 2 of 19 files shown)


scalvert:
This seems like a reversal of #612, which I don't think we should do.

That PR, and the work surrounding it, was explicitly done in an effort to incrementally move the `Task` interface to a simpler place. This makes the interface very simple (no nested props), and helps simplify things for task authors. Additionally, it imposes more 'clutter' in tasks rather than hiding that complexity elsewhere.

I'd prefer we instead pass the task instance to the builder function that requires it. 

Example:

Before

```ts
  async run(): Promise<Result[]> {
    let types = SEARCH_PATTERNS.flatMap((pattern) => {
      let files = this.context.paths.filterByGlob(pattern.pattern);
      return buildResultsFromPathArray(
        normalizePaths(files, this.context.cliFlags.cwd),
        pattern.patternName
      ).map((type) => this.appendCheckupProperties(type));
    });
    return types;
  }
```

After

```ts
  async run(): Promise<Result[]> {
    let types = SEARCH_PATTERNS.flatMap((pattern) => {
      let files = this.context.paths.filterByGlob(pattern.pattern);
      return buildResultsFromPathArray(this, 
        normalizePaths(files, this.context.cliFlags.cwd),
        pattern.patternName
      );
    return types;
  }
```

And for the builders, you can define a new, constrain type that the builder accepts:

```ts
Pick<Task, 'taskName' | 'taskDisplayName' | 'category' | 'group'>
```

This would then allow tasks to correctly build their results using all the data they need, and won't require any changes to the existing, simple `Task` interface.

Does this make sense?

carakessler:
I don't recall the context on #612 (ironic that this is #812), but I think passing `this` into a utility function that is cleanly abstracted away from `Task`s to be an antipattern. Why should these builders have an understanding of what is in a `Task`? 

My first preference would just be to bucket the data as done in this PR - I don't see how bucketing these under `TaskMetadata` is cluttered; if anything to me it is a logical grouping of concerns. Second would I suppose be to move the builders to `BaseTask` in lieu of passing `this` to the builders. 

Thoughts?

scalvert:
So my main goals here are
1. Keeping the Task interface simple (not nesting props)
1. Reducing the complexity of wrapping/unwrapping the properties (wrapping them only serves this one use case)
1. Passing the complexity to the builder function to execute the property merge in one pass, via this, which already encapsulates those properties (so why wrap them further?)

I'd prefer not to couple the builders with the base class, as I'm a big proponent of keeping base classes as lean as possible to maximize potential reuse (what happens if we want to extend from the base for something that doesn't require the builders?). Also, it makes the builders a bit more cluttered to test, as you'd now need to test them via the class instance vs. just as a plain function.

The builders are explicitly building a result from a Task. That's the association, and since they require data from the Task, there's natural coupling. This is not much different than your approach of using TaskMetadata, just without an extra layer of abstraction.

PR #830, opened 2020-12-23
Removing meta-task as a concept, moving logic from project-meta-task into get-log

carakessler:
There is a lot of infrastructure around meta-tasks that is not really being used. Given that there is consensus that it doesnt seem likely for there to be any additional metatasks, removed the infrastructure and moved the logic of the task into get-log.


scalvert:
@carakessler can you rebase with master?

scalvert (approved):

PR #831, opened 2020-12-23
    Removing empty Results from Result[]

carakessler:
Right now, when a task doesnt find any data for results, it creates sort of an empty Result with the message 'No results found' for a given task. This was a hack put in when we realized that Results couldn't have an occurrenceCount of 0. This doesnt really align with the principle of what SARIF spec wants - its basically just creating a Result with occurrenceCount = 0 without setting occurrenceCount... more confusing than bug in the first place.

This change removes that hack, and only returns actual/meaningful Results. In the verbose-console-reporter, added function that renders a list of the tasks without results found, so it will be clearer for humans consuming this data.

Now, tasks without Results will render like this: 
```
No Results Found For
■ Ember In-Repo Addons / Engines
■ Number of template-lint-disable Usages
```


scalvert (approved):
Thanks for this!

PR #847, opened 2021-01-14
Updates Octane migration task with better messages

scalvert:
The messages we receive in the SARIF log come directly from the eslint/ember-template-lint task outputs. While these are OK, they don't make as much sense in the context of Checkup, in that Check's goal is to aggregate activities and help direct them.

To help alleviate this, I've updated the Octane task such that we provide more Checkup specific messages, which are in the format of:

```
Octane | <FEATURE> : <FEATURE MESSAGE>. More info: <INFO URL>
```

This results in the following in the SARIF viewer:

Screen Shot 2021-01-14 at 10 51 52 AM

And when clicking through to the issue in the file:

Screen Shot 2021-01-14 at 10 52 21 AM


scalvert:
cc/ @chriskrycho @gabrielcsapo

carakessler (approved):
LGTM!

PR #848, opened 2021-01-14
Filtering out eslint results that indicate a rule was not found

carakessler:
When you run eslint parser on files that `eslint-disable` rules that are not defined  in the parser, it throws an error

An example would be error: `Definition for rule 'ember-best-practices/no-side-effect-cp' was not found.` being added from a file that has `/* eslint-enable ember-best-practices/no-side-effect-cp */`, when `ember-best-practices` rules are not being run in the Task

Filtered thought out of Results and added a `Total:` count in the default verbose console output

### DIFF
--- a/packages/core/src/data/builders.ts
+++ b/packages/core/src/data/builders.ts
@@ -55,6 +55,7 @@ function buildLocationDataFromPath(path: string): Location[] {
  * @param taskContext {Task} This is used to set Task properties on the Result
  * @param lintResults {LintResult[]} The LintResults used to create Result
  * @param [additionalData] {Object} Any additional data to be put into the properties bag
+ * @param [customMessages] {Record<string, string>} Custom messages to be rendered for each lintRule
  */
 export function buildResultsFromLintResult(
   taskContext: Pick<Task, 'taskName' | 'taskDisplayName' | 'category' | 'group'>,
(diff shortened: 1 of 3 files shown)


packages/core/src/data/builders.ts
```diff
@@ -65,23 +65,31 @@ export function buildResultsFromLintResult(
...
-      properties: {
-        ...{
-          taskDisplayName: taskContext.taskDisplayName,
-          category: taskContext.category,
-          group: taskContext.group,
+  return lintResults
+    .filter((lintResult) => {
+      // When files disable lint rules that are not defined in the context of our Task's parser, it adds a lintResult
+      // that indicates that the rule was not found. Since this is not an actual error || warning, we filter them out
+      const lintRuleNotDefinedRegex = new RegExp('Definition for rule .* was not found', 'g');
```
scalvert, line 72: I don't believe we need the 'g' flag on this regex, as you're just testing a single occurrence. You only need the 'g' flag if your planning on looping/testing multiple occurrences.

It's also actively bad if you test your string multiple times, and in particular when used with `test`. 

```js
let pattern = /a/g;
console.log(pattern.test('a')); // true
console.log(pattern.test('a')); // false
```

We should probably also create this RegExp in module scope, as it's not dynamic and the way its written now we'll create an instance for each loop.

scalvert (changes requested):

packages/core/src/data/builders.ts
carakessler: good call out, fixing

packages/core/src/data/builders.ts
```diff
@@ -55,6 +55,7 @@ function buildLocationDataFromPath(path: string): Location[] {
  * @param taskContext {Task} This is used to set Task properties on the Result
  * @param lintResults {LintResult[]} The LintResults used to create Result
  * @param [additionalData] {Object} Any additional data to be put into the properties bag
+ * @param [customMessages] {Record<string, string>} Custom messages to be rendered for each lintRule
```
scalvert, line 58: Thanks for adding this!

scalvert (approved):

PR #849, opened 2021-01-15
Only conditionally show total results in verbose-reporter

carakessler:
Only show total results summary if there is more than 1 bullet point (result) shown


packages/cli/src/reporters/verbose-console-reporter.ts
```diff
@@ -182,7 +182,11 @@ function getReportComponent(taskResults: Result[]) {
 
   ui.section(groupedTaskResults[0].properties?.taskDisplayName, () => {
     const totalResults = sumOccurrences(groupedTaskResults);
-    ui.log(`Total: ${totalResults}`);
+
+    // if there is only going to be 1 bullet point shown, we dont need to sum the total
```
scalvert, line 186: I don't know if we need this comment...

scalvert (approved):

scalvert:
@carakessler would you mind rebasing with master? Also there's an outstanding comment on a comment... :)

PR #854, opened 2021-01-22
Fix minor styling issues in plugin READMEs

tylerbecks:
Before:
Screen Shot 2021-01-22 at 5 52 35 PM

After:
Screen Shot 2021-01-22 at 5 51 51 PM


scalvert (approved):
Thank you!!

PR #859, opened 2021-01-29
Refactor and simplify builders

scalvert:
This PR works on cleaning up the builders so that consumers have a happier path to follow when authoring tasks.

It removes/refactors a number of the different builders that were cohabitating the `builders.ts` file. Builders are now split across two files:

- `sarif.ts` for SARIF-based builders
- `lint.ts` for lint-based builders (both `eslint` and `ember-template-lint`)

Consumers now access the builders by importing and invoking the associated builder's methods:

```
import { sarifBuilder } from '@checkup/core';

// ...

sarifBuilder.fromLintResults(...);
```


packages/core/src/data/path.ts
```diff
@@ -0,0 +1,7 @@
+export function normalizePath(path: string, cwd: string) {
+  return path.replace(`${cwd}/`, '');
```
rwjblue, line 2: Seems odd to replace cwd anywhere we find it in the string (I would expect this to be anchored at the beginning).

packages/core/src/data/path.ts
```diff
@@ -0,0 +1,7 @@
+export function normalizePath(path: string, cwd: string) {
```
rwjblue, line 1: I wouldn't call this "normalizePath" FWIW, seems more like "strip cwd" or something

rwjblue (commented):
Looks good overall, definitely seems like a nice improvement.

packages/core/src/utils/sarif-utils.ts
```diff
@@ -3,3 +3,15 @@ import { Result } from 'sarif';
 export function sumOccurrences(results: Result[]) {
   return results.reduce((total, result) => total + (result.occurrenceCount || 0), 0);
 }
+
+/*
+ * This function takes in multiple Results with different locations but the same message,
+ * and return a single Result to be rendered in the console (with occurrenceCounts added together)
+ */
+export function reduceResults(groupedResults: Result[][]): Result[] {
```
carakessler, line 11: this is the only naming that i still find a little too vague... i know you hate long names, but maybe `reduceResultsByMessage`? something that makes it clear in the name what we are reducing it by? especially since we may at some point what to reduce/group results based on another shared property

carakessler (approved):

packages/core/src/data/path.ts
scalvert: It is. We're using the [`string` variant](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace), not the `RegExp` one.

packages/core/src/data/path.ts
scalvert: Yep, seems more specific.

packages/core/src/utils/sarif-utils.ts
scalvert: It's not doing it by message though, is it?

PR #864, opened 2021-02-08
Fixing task config overrides merging correctly

scalvert:
The `deepmerge` package used to facilitate the deep merge of task configs overriding eslint or ember-template-lint configurations for a task was incorrectly merging tuple options. When it encountered two tuples, since they are arrays it would attempt to concat them, resulting in an incorrect structure.

## Example of incorrect merge

Linting config:

```js
{
  rules: {
    'fake-eslint-rule': [
      'warn',
      {
         prop1: 'error',
         prop2: 'off',
      },
    ],
 },
}
```

Task config:

```js
{
  rules: {
    'fake-eslint-rule': [
      'error',
      {
         prop1: 'warn',
         prop2: 'off',
      },
    ],
 },
}
```

Merged configs:

```js
{
  rules: {
    'fake-eslint-rule': [
      'warn',
      {
         prop1: 'error',
         prop2: 'off',
      },
      'error',
      {
         prop1: 'warn',
         prop2: 'off',
      },
    ],
 },
}
```

## Example of correct merge

```js
{
  rules: {
    'fake-eslint-rule': [
      'error',
      {
         prop1: 'warn',
         prop2: 'off',
      },
    ],
 },
}
```

### DIFF
--- a/packages/core/__tests__/utils/merge-lint-config-test.ts
+++ b/packages/core/__tests__/utils/merge-lint-config-test.ts
@@ -1,0 +1,217 @@
+import { mergeLintConfig } from '../../src/utils/merge-lint-config';
+import { TemplateLintConfig } from '../../src/types/ember-template-lint';
+import { CLIEngine } from 'eslint';
+
+describe('mergeLintConfig', () => {
+  describe('eslint', () => {
+    it('should merge strings and tuples when strings are leftmost, tuple takes precedence', () => {
+      let original: CLIEngine.Options = {
+        rules: {
+          'fake-eslint-rule': 'error',
+        },
+      };
+
+      let overrides = {
+        rules: {
+          'fake-eslint-rule': [
+            'warn',
+            {
+              prop1: 'error',
+              prop2: 'off',
+            },
+          ],
+        },
+      };
+
+      expect(mergeLintConfig(original, overrides)).toMatchInlineSnapshot(`
+        Object {
+          "rules": Object {
+            "fake-eslint-rule": Array [
+              "warn",
+              Object {
+                "prop1": "error",
+                "prop2": "off",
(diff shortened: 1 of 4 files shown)


scalvert:
cc/ @chriskrycho for ensuring our test cases cover the bug you experienced.

chriskrycho (changes requested):
The cases as covered look good. The one case we're missing is, I think, the case we want to handle for the motivating scenario: where the source has a given rule setting and config, and the override has the same *setting* but different *config*. I think the implementation handles this correctly, based on what I see with the tuple merging with `"warn"` and `"error"` but the test as written would equally work with the right side simply smashing the left as with merging the rule tuple.

Here's a test case which should catch the difference between a smash and a merge:

```js
    it('should merge rule tuples when there are rule overrides', () => {
      let original: CLIEngine.Options = {
        rules: {
          'fake-eslint-rule': [
            'error',
            {
              prop1: 'error',
              prop2: 'off',
            },
          ],
        },
      };

      let overrides = {
        rules: {
          'fake-eslint-rule': [
            'error',
            {
              prop1: 'warn',
              prop3: 'error',
            },
          ],
        },
      };

      expect(mergeLintConfig(original, overrides)).toMatchInlineSnapshot(`
        Object {
          "rules": Object {
            "fake-eslint-rule": Array [
              "error",
              Object {
                "prop1": "warn",
                "prop2": "off",
                "prop3": "error",
              },
            ],
          },
        }
      `);
    });
```

This test case requires the merge behavior:

- it uses `prop1` right the right
- it preserves `prop2` from the left
- it incorporates `prop3` from the right

chriskrycho (approved):
🎉 :shipit: thank you!

PR #881, opened 2021-02-18
Removing `output-map` from verbose-console-reporter

carakessler:
This is the first in a series of refactors to verbose-console-reporter. There will also be a change where some additional util functions (similar to `renderLintingSummaryResult`) exposed to offer an additional layer of options between using ui library and building result per task directly, and just using the current default behavior.

This change simply removes `output-map` and all task specific code from verbose-console-reporter, and moves it to the plugins directly. Our internal plugins should register custom reporters the same way we would expect an external plugin to. There are no functionality changes in this PR, just moving existing custom cases (where necessary) to plugins

### DIFF
--- a/packages/checkup-plugin-ember/package.json
+++ b/packages/checkup-plugin-ember/package.json
@@ -42,7 +42,8 @@
     "hooks": {
       "register-info-tasks": "./lib/hooks/register-info-tasks",
       "register-migration-tasks": "./lib/hooks/register-migration-tasks",
-      "register-actions": "./lib/hooks/register-actions"
+      "register-actions": "./lib/hooks/register-actions",
+      "register-task-reporter": "./lib/hooks/register-task-reporters"
     }
   },
   "publishConfig": {
--- a/packages/core/src/utils/verbose-reporter-utils.ts
+++ b/packages/core/src/utils/verbose-reporter-utils.ts
@@ -1,0 +1,38 @@
+import { NO_RESULTS_FOUND } from '../data/sarif';
+import { groupDataByField } from '../data/formatters';
+import { reduceResults, sumOccurrences } from './sarif-utils';
+import { ui } from './ui';
+import { Result } from 'sarif';
+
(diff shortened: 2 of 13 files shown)


packages/checkup-plugin-ember/package.json
```diff
@@ -42,7 +42,8 @@
     "hooks": {
       "register-info-tasks": "./lib/hooks/register-info-tasks",
       "register-migration-tasks": "./lib/hooks/register-migration-tasks",
-      "register-actions": "./lib/hooks/register-actions"
+      "register-actions": "./lib/hooks/register-actions",
+      "register-task-reporter": "./lib/hooks/register-task-reporters"
```
carakessler, line 46: this hook should be called `register-task-reporters` not `register-task-reporter`. Will update in a separate subsequent PR

scalvert: Correct. I messed this up, and noticed it last week when I was working on adding one. Thanks for looking at fixing.

packages/core/src/utils/verbose-reporter-utils.ts
```diff
@@ -0,0 +1,38 @@
+import { NO_RESULTS_FOUND } from '../data/sarif';
```
scalvert, line 1: I think this is fine for now. Ultimately, I'd like us to have a set of components that folks can use, and you would specify how your result is output via task config in the SARIF log. Because we're leveraging SARIF's Result type, we can predictably render things, as they all stem from the same data schema now. Providing different components for different output is ultimately what we wanted to produce a year ago, but now have a chance to do it.

scalvert (approved):

packages/core/src/utils/verbose-reporter-utils.ts
carakessler: Yeah definitely, as it says in description there will be subsequent PRs that contain more significant refactors. This one is pretty much just movin code around

PR #882, opened 2021-02-18
Populating `Result.kind` field

carakessler:
This field is used to specify the nature of the Results.

For `Results` being built `fromLintResults`, I set default as `fail` for `Result.kind`.
For `Results` that represent the fact that no results were found, I set `Result.kind` as `notApplicable`
For all other `Results`, default for `Result.kind` is `informational`. This can be optionally overridden at the Task level

### DIFF
--- a/packages/core/src/data/sarif.ts
+++ b/packages/core/src/data/sarif.ts
@@ -8,16 +8,18 @@ type SarifTaskIdentifier = Pick<Task, 'taskName' | 'taskDisplayName' | 'category
 /**
  * Builds SARIF Results from a list of LintResults.
  *
- * @param taskContext {Task} This is used to set Task properties on the Result
- * @param lintResults {LintResult[]} The LintResults used to create Result
- * @param [additionalData] {Object} Any additional data to be put into the properties bag
- * @param [customMessages] {Record<string, string>} Custom messages to be rendered for each lintRule
+ * @param {SarifTaskIdentifier} taskContext  This is used to set Task properties on the Results
+ * @param {LintResult[]} lintResults  The LintResults used to create Results
+ * @param {Object} [additionalData] Any additional data to be put into the properties bag
+ * @param {Record<string, string>} [customMessages]  Custom messages to be rendered for each lintRule
+ * @param {Result.kind} [resultKind='fail']  This is used to specify the nature of the Results
  */
 function fromLintResults(
   taskContext: SarifTaskIdentifier,
   lintResults: LintResult[],
-  additionalData: object = {},
-  customMessages: Record<string, string> = {}
+  additionalData: Object = {},
+  customMessages: Record<string, string> = {},
+  resultKind: Result.kind = 'fail'
 ): Result[] {
   if (lintResults.length === 0) {
     return buildEmptyResult(taskContext);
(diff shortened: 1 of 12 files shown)


packages/core/src/data/sarif.ts
```diff
@@ -56,21 +59,24 @@ function fromLintResults(
...
+ * @param {string[]} locations The paths used to create Results
+ * @param {string} message The message that identifies the data represented in the Results
+ * @param {Result.kind} [resultKind='informational']  This is used to specify the nature of the Results
  */
 function fromLocations(
   taskContext: SarifTaskIdentifier,
   locations: string[],
-  message: string
+  message: string,
+  resultKind: Result.kind = 'informational'
```
gabrielcsapo, line 71: should the default be unknown?

packages/core/src/data/sarif.ts
```diff
@@ -87,17 +93,24 @@ function fromLocations(
...
+ * @param {T[]} data  The raw data used to derive the Result's count
+ * @param {string} message The message that identifies the data represented in the Results
+ * @param {Result.kind} [resultKind='informational']  This is used to specify the nature of the Results
  */
-function fromData<T>(taskContext: SarifTaskIdentifier, data: T[], message: string): Result[] {
+function fromData<T>(
+  taskContext: SarifTaskIdentifier,
+  data: T[],
+  message: string,
+  resultKind: Result.kind = 'informational'
```
gabrielcsapo, line 105: same thought here about it being unknown.

packages/core/src/data/sarif.ts
carakessler: `Result.kind` is an enum, and unknown is not an option.

packages/core/src/data/sarif.ts
carakessler: `Result.kind` is an enum, and unknown is not an option.

packages/core/src/data/sarif.ts
```diff
@@ -8,16 +8,18 @@ type SarifTaskIdentifier = Pick<Task, 'taskName' | 'taskDisplayName' | 'category
...
+ * @param {Result.kind} [resultKind='fail']  This is used to specify the nature of the Results
  */
 function fromLintResults(
   taskContext: SarifTaskIdentifier,
   lintResults: LintResult[],
-  additionalData: object = {},
-  customMessages: Record<string, string> = {}
+  additionalData: Object = {},
+  customMessages: Record<string, string> = {},
+  resultKind: Result.kind = 'fail'
```
rwjblue, line 22: Having this many optional arguments sticks out like a bit of a sore thumb. Callers are going ot have to either pad with `undefined` for `additionalData`/`customMessages` to be able to specify a `resultKind` (you can see that up in packages/checkup-plugin-ember/src/tasks/ember-test-types-task.ts).

scalvert: Agree. My plan was to refactor those so that the signatures are a lot simpler, but I was hoping to do something like #883 first, and build up from there. Ultimately, it's much easier to pass in an options object that can be updated/changed without having to continually widen the signature.

scalvert:
Closing in favor of building out the SARIF result builder.

PR #883, opened 2021-02-19
Adding low-level SARIF result builder

scalvert:
Creates a simple, low-level SARIF `Result` object builder. 

In addition, it also introduces a `SarifLogBuilder` class, which we're ultimately going to need if we want to provide the capability of annotating the SARIF log with extra metadata. Simply returning a `Result[]` from the Task won't be enough if we want to be able to cross reference metadata in `rules`, `locations`, or the like. 

We'll likely need to instantiate a single `SarifLogBuilder` instance for each command instance, pass it through to the Task classes, and allow them to build the SARIF log this way. If they want to add extra data to the log, they can access it directly via `new SarifLogBuilder().log`.

### DIFF
--- a/packages/core/src/data/sarif-log-builder.ts
+++ b/packages/core/src/data/sarif-log-builder.ts
@@ -1,19 +1,36 @@
-import { Log, Run, ReportingDescriptor, Result } from 'sarif';
-import { builder } from './sarif';
+import {
+  Log,
+  Run,
+  ReportingDescriptor,
+  Result,
+  Address,
+  Artifact,
+  ArtifactLocation,
+  Conversion,
+  ExternalPropertyFileReferences,
+  Graph,
+  Invocation,
+  LogicalLocation,
+  PropertyBag,
+  RunAutomationDetails,
+  SpecialLocations,
+  ThreadFlowLocation,
+  Tool,
+  ToolComponent,
+  VersionControlDetails,
+  WebRequest,
+  WebResponse,
+} from 'sarif';
+import ow from 'ow';
+import { SetRequired } from 'type-fest';
 
-const RUN_DEFAULTS = {
-  tool: {
-    driver: {
-      name: 'checkup',
-      language: 'en-US',
-      informationUri: 'https://github.com/checkupjs/checkup',
-    },
-  },
-};
+type RequiredRun = SetRequired<Run, 'tool' | 'results'>;
+type RequiredResult = SetRequired<Result, 'message' | 'ruleId' | 'kind' | 'level'>;
 
 export class SarifLogBuilder {
   log: Log;
   runs: Run[];
+  currentRun!: RequiredRun;
   rules: ReportingDescriptor[];
   results: Result[];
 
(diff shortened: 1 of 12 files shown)


packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,58 @@
...
+        results: this.results,
+      },
+      ...run,
+    });
+  }
+
+  addRule(rule: ReportingDescriptor) {
+    if (!this.rules.find((r) => r.id === rule.id)) {
+      this.rules.push(rule);
+    }
```
rwjblue, line 52: Can we make it a Map instead?

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  addResult(result: Result) {
+    this.results.push(builder.buildResult(result));
```
rwjblue, line 56: Maybe rename `builder` -> `sarifBuilder`

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,58 @@
...
+    this.runs = [];
+    this.rules = [];
+    this.results = [];
+    this.log = {
+      version: '2.1.0',
+      $schema: 'https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json',
+      runs: this.runs,
+    };
+
+    this.addRun();
```
rwjblue, line 30: Why do we immediately call addRun?

scalvert:
Moving this work over to https://github.com/microsoft/sarif-js-sdk

packages/core/src/data/sarif-log-builder.ts
scalvert: Updated the comment to reflect. We always expect there to be at least one run, so we pre-populate it.

packages/core/src/data/sarif-log-builder.ts
scalvert: This is gone now.

packages/core/src/data/sarif-log-builder.ts
scalvert: This has been updated, though we still don't use a Map. We're simply using the rules array from the `ToolComponent`, without keeping a local version around. This makes it easier to manage since you could have _n_ runs.

packages/core/src/data/sarif-log-builder.ts
rwjblue: Is there ever a case where folks need to customize it though? In that case, they'd have to do `builder.runs[0]` to mutate the implicitly created one. Maybe thats uncommon (and therefore totally fine to be less ergonomic), I just wanted to double check with you.

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+
+type RequiredRun = SetRequired<Run, 'tool' | 'results'>;
+type RequiredResult = SetRequired<Result, 'message' | 'ruleId' | 'kind' | 'level'>;
+
+export class SarifLogBuilder {
+  log: Log;
+  runs: Run[];
+  currentRun!: RequiredRun;
+  rules: ReportingDescriptor[];
+  results: Result[];
```
rwjblue, line 35: Are these all public? I _think_ they are, right? Should we make that explicit?

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+  tool: Tool;
+  threadFlowLocations?: ThreadFlowLocation[] | undefined;
+  translations?: ToolComponent[] | undefined;
+  versionControlProvenance?: VersionControlDetails[] | undefined;
+  webRequests?: WebRequest[] | undefined;
+  webResponses?: WebResponse[] | undefined;
+  properties?: PropertyBag | undefined;
+
+  constructor() {
+    this.tool = {
```
rwjblue, line 142: We're a tool

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    this.currentRun.results.push(result);
+  }
+
+  toJson() {
+    return JSON.stringify(this.log, null, 2);
+  }
+}
+
+class RunBuilder implements RequiredRun {
```
rwjblue, line 111: Just curious why is this a class?

packages/core/src/data/sarif.ts
```diff
@@ -5,6 +5,19 @@ export const NO_RESULTS_FOUND = 'No results found';
 
 type SarifTaskIdentifier = Pick<Task, 'taskName' | 'taskDisplayName' | 'category' | 'group'>;
 
+/**
+ *
+ *
+ * @param result Builds a SARIF Result
+ */
+function buildResult(result: Result): Result {
+  if (!result.message) {
```
rwjblue, line 14: ow?

packages/core/src/data/sarif-log-builder.ts
scalvert: Yep, they're all public. We _can_ make them explicit, but the only ones I usually mark are `private` fields. I'll leave for now, but we can look to update when we move this to `@microsoft/sarif-js-sdk`.

packages/core/src/data/sarif-log-builder.ts
scalvert: I was envisioning having this be the class that's passed around inside Tasks. The reason for this is that it more narrowly encapsulates the functionality that Tasks need (results), while not allowing for access to the rest of the log.

packages/core/src/data/sarif.ts
scalvert: I think this is going to go in favor of the actual methods in the builder class.

packages/core/src/data/sarif-log-builder.ts
scalvert: You are :)

scalvert:
https://github.com/checkupjs/checkup/pull/883#discussion_r639842051

True, and fair point. I can change this.

PR #886, opened 2021-02-19
bugfix: Fix checkup versions in plugin

chadhietala:
This fixes an issue where the checkup versions are stale when stamping
out a new plugin leaving the developer with a completely broken setup.


scalvert:
Thanks for tackling this, @chadhietala. It's been on my list for a while, but I keep putting it off.

scalvert (approved):

PR #887, opened 2021-02-19
Merging info and migration commands

scalvert:
After extensive discussions with @carakessler and @rwjblue, we're merging the `info` and `migration` commands back together. This keeps tasks that are not destined for the `validate` command under the same invocation context. 

This is the first step in a series of improvements/simplifications for v1.0.0.


rwjblue (changes requested):
I can't comment on it (likely due to no content) but I think you need to remove hte `.DS_Store` file (and probably add it to your global gitignore too).

rwjblue (commented):
Other than that LGTM

scalvert:
Weird. Not sure why that made it in. Adding it to my global.

Issue #866, opened 2021-02-12
Pull out dependency on oclif 


carakessler:
Checkup seems to have outgrown oclif, and its gotten to the point where it is causing more problems than it is solving, particularly when it comes to the flexibility of the node API. 

We need to remove our dependency on oclif, and build out the hooks, config, command, error, and flags infrastructure ourselves. 

- [x] Create parallel `checkup` bin script
- [x] Create top level command, sub commands (`run` and `validate`)
- [x] Parse arguments and ensure correct usage output
- [x] Decouple command classes from oclif
- [x] Implement plugin loading

scalvert:
Implemented in #900

PR #891, opened 2021-03-02
Part 1 of moving off oclif to vanilla CLI

scalvert:
This is part 1 of a multi-part refactor to address #866.

This PR

- creates new bin script
- creates new Checkup class to encapsulate bin functionality
- creates new `RunCommand` and `GenerateCommand` classes to encapsulate the two commands

Once merged, the functionality of the existing commands can be migrated over to these new commands.


packages/cli/bin/checkup
```diff
@@ -0,0 +1,7 @@
+require('v8-compile-cache');
```
rwjblue, line 1: I would normally have this file end with `.js`, any reason you didn't?

packages/cli/src/api/generate.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default class GenerateCommand {
+  options: GenerateOptions;
+
+  constructor(options: yargs.Arguments) {
+    this.options = options as GenerateOptions;
+  }
+
+  async run() {
+    console.log(this.options);
```
rwjblue, line 19: Do you plan to land this as is (with `console.log` here), or is this more of a draft / proof of concept?

packages/cli/src/api/run.ts
```diff
@@ -0,0 +1,34 @@
+import * as yargs from 'yargs';
+
+export type RunOptions = {
+  version: void;
+  help: void;
+  config: string | undefined;
```
rwjblue, line 6: why `| undefined` vs `config?:`? (here and the ones below)

packages/cli/src/api/run.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.options = options as RunOptions;
+  }
+
+  async run() {
+    await new Promise((resolve) => {
+      setTimeout(resolve, 2000);
+    });
+
+    return {
+      message: 'I ran',
```
rwjblue, line 31: 🏃‍♂️

packages/cli/src/checkup.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import RunCommand from './api/run';
+import GenerateCommand from './api/generate';
+
+async function parseArgv(argv: string[]) {
+  let parser = yargs
+    .scriptName('checkup')
+    .usage(
+      `
+A health checkup for your project
```
rwjblue, line 13:
```suggestion
A health checkup for your project ✅
```

packages/cli/src/checkup.ts
```diff
@@ -0,0 +1,150 @@
...
+            'list-tasks': {
+              alias: 'l',
+              description: 'List all available tasks to run.',
+              boolean: true,
+            },
+          });
+      },
+      handler: async (argv: yargs.Arguments) => {
+        let cmd = new RunCommand(argv);
+        let spinner = ora().start('Checking up on your project');
```
rwjblue, line 91:
```suggestion
        let spinner = ora().start('🕵️‍♀️ Checking up on your project');
```

packages/cli/src/checkup.ts
```diff
@@ -0,0 +1,150 @@
...
+              default: '.',
+              description: 'The path referring to the directory that the generator will run in',
+            },
+          });
+      },
+      handler: async (argv: yargs.Arguments) => {
+        let cmd = new GenerateCommand(argv);
+
+        await cmd.run();
+      },
```
rwjblue, line 134: Should this also use ora?

packages/cli/src/checkup.ts
```diff
@@ -0,0 +1,150 @@
...
+    })
+    .wrap(yargs.terminalWidth())
+    .help()
+    .version();
+
+  if (argv.length === 0) {
+    parser.showHelp();
+    parser.exit(1, new Error('Showing help'));
```
rwjblue, line 142: Why do you have to pass an error here?

packages/cli/bin/checkup
scalvert: Honestly no reason at all. I can update if you feel strongly, I don't really.

packages/cli/src/api/generate.ts
scalvert: Ya this is part of a multi-part, parallel set of changes that I'd like to land in increments. The plan is to do the cli infra, layer in the command functionality, and all the other bits and bobs separately. So, this is intended to merge as is, and this call will be replaced with the real functionality.

packages/cli/src/api/run.ts
scalvert: Potato, potAto...

packages/cli/src/api/run.ts
scalvert: I ran so far away... 🐦

packages/cli/src/checkup.ts
scalvert: So ora has know issues when interacting with inquirer, which yeoman uses. The generate command defers almost immediately to yeoman, which uses prompts to generate the specific functionality. I think using ora here doesn't make as much sense. WDYT?

packages/cli/src/checkup.ts
scalvert: The types in @types/yargs required it.

packages/cli/bin/checkup
rwjblue: Without .js or .ts it is more susceptible to fundamental issues because it won't get linted

packages/cli/src/api/generate.ts
rwjblue: Sounds good.

packages/cli/src/api/run.ts
rwjblue: Sure, but we shouldn't have both potato and tomato in the same PR 😜

packages/cli/src/checkup.ts
rwjblue: Ya sounds good. Maybe add a comment (for future travelers)

packages/cli/src/checkup.ts
rwjblue: 👊

packages/cli/bin/checkup
scalvert: Updated.

packages/cli/src/api/run.ts
scalvert: Updated the potatoes.

rwjblue (approved):
Not sure when you intend to land each piece, but this looks good to me as of what is pushed right now.

packages/cli/src/checkup.ts
scalvert: I've had to remove it because it causes an unexpected extra error message in the usage output. It's annoying that this type is forcing me to provide this param when it's not needed.

packages/cli/src/checkup.ts
scalvert: Good call.

scalvert:
@rwjblue I'm just going to land each as they are done. Since I'm implementing them in parallel to the existing functionality, I think that'll work nicely as I move functionality over incrementally. Once they're all done, I'll delete the old functionality.

PR #895, opened 2021-03-05
Part 2 of moving off oclif to vanilla CLI

scalvert:
This second part moves the generator command to the new CLI. All other generator features are kept, except the `force` option, which actually did nothing...weeeeeeee!

I've also reversed some of the task generator work that was done to support multiple commands, which has since been reversed.

The new generators CLI options are more specific than they used to be. Now we actually create a separate sub-sub command for each generator.

```bash
checkup generate

Runs a generator to scaffold Checkup code

Commands:
  checkup g plugin <name> [options]   Generates a checkup plugin project
  checkup g task <name> [options]     Generates a checkup task within a project
  checkup g actions <name> [options]  Generates checkup actions within a project
  checkup g config                    Generates a .checkuprc within a project

Options:
  --help     Show help                                                         [boolean]
  --version  Show version number                                               [boolean]
```

This allows us to have a much better overall experience when interacting with generators. As an added bonus, I also added context specific validation to ensure generators are only run in the correct context.


packages/cli/src/api/checkup-task-runner.ts
```diff
@@ -0,0 +1,26 @@
+export type RunOptions = {
+  cwd: string;
+  config?: string;
+  category?: string[];
+  group?: string[];
+  task?: string[];
```
carakessler, line 6: should this be `tasks`? i know the cli has worked like `--task foo-task --task bar-task` but since we are passing in an array of strings with task names, `tasks` makes more sense to me

packages/cli/src/api/checkup-task-runner.ts
```diff
@@ -0,0 +1,26 @@
...
+  excludePaths?: string[];
+};
```
carakessler, line 8: what about `paths`? can we expose a way for that to also be passed into node api?

packages/cli/src/api/checkup-task-runner.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  async run() {
+    await new Promise((resolve) => {
+      setTimeout(resolve, 2000);
+    });
+
+    return {
+      message: 'I ran',
+    };
+  }
```
carakessler, line 25: is this meant to be submitted?

packages/cli/src/extended-error.ts
```diff
@@ -0,0 +1,19 @@
...
+  error: string | Error | CheckupError,
+  options: { code?: string; exit?: number | false } = {}
+) {
+  if (typeof error === 'string') {
+    error = new Error(error);
+  }
+
+  if (options.exit === false) {
+    console.error(
+      error instanceof CheckupError ? error.render() : `\n${red('Error')} ${error.message}`
```
rwjblue, line 14: the else case of this ternary seems vaguely odd to me

I would expect something like:

```suggestion
      error instanceof CheckupError ? error.render() : error
```

(console.error is going to "do the right thing" with the actual error itself, we shouldn't pass it a string like that)

packages/cli/src/api/checkup-task-runner.ts
scalvert: Ya I think it could be. Ultimately we allow multiple, so single task flags get merged into the array. I could for sure also apply to the others. I think in this case, since it's internal, we should probably pluralize.

packages/cli/src/api/checkup-task-runner.ts
scalvert: Yep!

packages/cli/src/api/checkup-task-runner.ts
scalvert: Hehe. I ran so far away...

packages/cli/__tests__/cli-test.ts
```diff
@@ -85,14 +85,13 @@ describe('cli-test', () => {
       Runs a generator to scaffold Checkup code
 
       Positionals:
-        type  Type of generator to run  [choices: \\"config\\", \\"plugin\\", \\"task\\", \\"actions\\"]
-        name  Name of the entity (kebab-case)  [required]
+        generator  Type of generator to run  [required] [choices: \\"config\\", \\"plugin\\", \\"task\\", \\"actions\\"]
+        name       Name of the entity (kebab-case)  [required] [default: \\"\\"]
```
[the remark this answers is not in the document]
scalvert: So there's 4 generators supported. 3 of them require name, 1 does not (the config generator). I'm also not sure how to mark a positional as optional, or if that's weird.

packages/cli/src/api/generator.ts
```diff
@@ -0,0 +1,79 @@
...
+import { existsSync, rmdirSync } from 'fs-extra';
+
+export type GenerateOptions = {
+  generator: string;
+  name: string;
+  path: string;
+  defaults?: boolean;
+};
+
+const VALID_GENERATORS = ['config', 'plugin', 'task', 'actions'];
```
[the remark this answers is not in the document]
scalvert: Ya makes sense.

packages/cli/src/api/generator.ts
```diff
@@ -0,0 +1,79 @@
...
+  async generate(type: string, generatorOptions: GenerateOptions) {
+    this.debug('generatorOptions', generatorOptions);
+
+    const env = createEnv();
+
+    env.register(require.resolve(`../generators/${type}`), `checkup:${type}`);
+
+    try {
+      await new Promise((resolve, reject) => {
+        env.run(`checkup:${type}`, generatorOptions, (err: Error | null) => {
```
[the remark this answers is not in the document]
scalvert: `env.run` already has a promise version. I just didn't want to change it in this PR, but will convert to it once this merges. I use the promise version in ember-template-lint.

packages/cli/src/checkup.ts
```diff
@@ -87,7 +87,15 @@ checkup <command> [options]`
...
       },
       handler: async (argv: yargs.Arguments) => {
-        let cmd = new RunCommand(argv);
+        let cmd = new CheckupTaskRunner({
+          excludePaths: argv.excludePaths as string[],
+          config: argv.config as string,
+          cwd: argv.cwd as string,
+          category: argv.category as string[],
+          group: argv.group as string[],
+          task: argv.task as string[],
```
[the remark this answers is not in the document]
scalvert: Yep, I was planning on narrowing these types after I got this working.

packages/cli/src/checkup.ts
```diff
@@ -101,25 +109,20 @@ checkup <command> [options]`
...
-          .positional('type', {
+          .positional('generator', {
             description: 'Type of generator to run',
             choices: ['config', 'plugin', 'task', 'actions'],
-            required: true,
           })
           .positional('name', {
             description: 'Name of the entity (kebab-case)',
-            require: false,
+            default: '',
```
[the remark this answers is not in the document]
scalvert: Name is only used for 3 of the generators. The config generator doesn't require a name.

packages/cli/src/extended-error.ts
scalvert: This is printing custom error messages. This is loosely related to how oclif prints its errors, and this function was originally mimicking what they do in terms of output for general errors, while adding an extra output for checkup specific errors.

packages/cli/src/api/checkup-task-runner.ts
scalvert: Done.

packages/cli/src/api/checkup-task-runner.ts
scalvert: Updated with a comment instead.

packages/cli/__tests__/cli-test.ts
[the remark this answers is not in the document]
scalvert: I've changed how this works. Instead, I've created a set of subcommands for each generator under the parent generate command. This has resulted in much cleaner setup of commands, and better usage output.

packages/cli/src/checkup.ts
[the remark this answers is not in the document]
scalvert: This is removed now.

packages/cli/src/extended-error.ts
scalvert: I'm going to leave this for now, and update at the end of the refactor once I get everything working. I'll likely just remove the custom checkup error.

rwjblue (approved):

packages/checkup-plugin-javascript/__tests__/__snapshots__/outdated-dependencies-task-test.ts.snap
```diff
@@ -36,7 +36,7 @@ Array [
         Object {
           "homepage": "https://ember-cli.com/",
           "installed": "3.20.0",
-          "latest": "3.25.1",
+          "latest": "3.25.2",
```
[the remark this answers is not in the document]
scalvert: Yes, it is. it's something we should absolutely change.

PR #903, opened 2021-03-26
Include `filePath` as part of `ember-template-lint-parser`, update `ember-template-lint`

brendenpalmer:
This PR ensures that we pass `filePath` to `ember-template-lint` as part of the built in `ember-template-lint-parser`; we also drop `moduleId` since it was deprecated as part of https://github.com/ember-template-lint/ember-template-lint/pull/1760. This also updates `ember-template-lint` to latest.

### DIFF
--- a/packages/core/src/parsers/ember-template-lint-parser.ts
+++ b/packages/core/src/parsers/ember-template-lint-parser.ts
@@ -27,20 +27,21 @@ class EmberTemplateLintParser implements Parser<TemplateLintReport> {
       template: fs.readFileSync(path, { encoding: 'utf8' }),
     }));
 
-    let results: TemplateLintResult[] = sources.map(({ path, template }) => {
-      let messages: TemplateLintMessage[] = this.engine.verify({
-        source: template,
-        moduleId: path,
-        filePath: path,
-      });
+    let results: TemplateLintResult[] = await Promise.all(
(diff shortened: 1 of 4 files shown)


packages/core/src/parsers/ember-template-lint-parser.ts
```diff
@@ -31,6 +31,7 @@ class EmberTemplateLintParser implements Parser<TemplateLintReport> {
       let messages: TemplateLintMessage[] = this.engine.verify({
         source: template,
         moduleId: path,
```
scalvert, line 33: Depending on the version of ember-template-lint we're using, we should consider removing `moduleId`, since it's [deprecated](https://github.com/ember-template-lint/ember-template-lint/pull/1760). Could you verify the version number we're using? We may want to bump the version simultaneously as this change lands.

scalvert (changes requested):

packages/core/src/parsers/ember-template-lint-parser.ts
brendenpalmer: checkup is currently using `ember-template-lint` version `2.18.1` right now; i've updated to latest and also updated `ember-template-lint-parser` to account for its new async `verify` API

scalvert:
Thanks for fixing this, @brendenpalmer!

PR #906, opened 2021-03-29
Fixes verbose output when using generator utils in tests

scalvert:
When using the generator utils in tests, there was some excessively verbose output indicating that the `group` question didn't have an answer provided, mainly due to our defaults not including a default for that property. This PR adds defaults for both `category` and `group` questions.


scalvert:
Adding you two as reviewers, but I'll likely just merge since this is just output cleanup.

PR #907, opened 2021-03-30
Speeds up tests by using local link to @checkup/core in generated plugins

scalvert:
Uses a symlink to the @checkup/core directory in the project during tests.

### DIFF
--- a/packages/cli/__tests__/__utils__/fake-project.ts
+++ b/packages/cli/__tests__/__utils__/fake-project.ts
@@ -1,15 +1,19 @@
-import { join } from 'path';
+import { join, resolve } from 'path';
 import * as helpers from 'yeoman-test';
 
 import { CheckupProject } from '@checkup/test-helpers';
 import { generatePlugin, generateTask } from './generator-utils';
 
 import type { Answers } from 'inquirer';
+import { mkdirp, symlink } from 'fs-extra';
 
 export class FakeProject extends CheckupProject {
   async addPlugin(options: helpers.Dictionary<any> = {}, prompts: Answers = {}) {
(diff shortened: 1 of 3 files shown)


packages/cli/__tests__/__utils__/fake-project.ts
```diff
@@ -1,15 +1,22 @@
...
   async addPlugin(options: helpers.Dictionary<any> = {}, prompts: Answers = {}) {
     let pluginDir = await generatePlugin(options, prompts, join(this.baseDir, 'node_modules'));
-    this.install(pluginDir);
+    let coreDir = join(pluginDir, 'node_modules', '@checkup', 'core');
+
+    // we create a self-referential link to the core package within the generated plugin. This
+    // allows us to generate plugins, and test the APIs for the latest version of core that is
+    // referenced via the plugins' node_modules.
+    await mkdirp(dirname(coreDir));
+    await symlink(join(__dirname, '../../..', 'core'), coreDir);
```
rwjblue, line 19: Can you rename this variable? Maybe `targetCoreDir`? Basically I would prefer:

```suggestion
    let source = join(__dirname, '../../..', 'core');
    let target = join(pluginDir, 'node_modules', '@checkup', 'core');

    // we create a self-referential link to the core package within the generated plugin. This
    // allows us to generate plugins, and test the APIs for the latest version of core that is
    // referenced via the plugins' node_modules.
    await mkdirp(dirname(target));
    await symlink(source, target);
```

scalvert: Yep good call.

rwjblue (approved):

PR #909, opened 2021-03-31
Moving registration to index

scalvert:
Completes the last step of refactoring as part of the move off oclif. In this PR, we're further simplifying the dependency registration in plugins. We move away from the vestiges of registration 'hook' files, and use the standardized `index.js` entry point to provide a `register` function, which is then invoked by Checkup to load and register plugin dependencies.

The register function in index.ts simply looks like this:

```ts
function register(args: RegistrationArgs): void;
```

And the types governing the registration:

```ts
export interface RegistrationArgs {
  context: TaskContext;
  register: RegistrationProvider;
}

export interface RegistrationProvider {
  actions(taskName: TaskName, evaluate: TaskActionsEvaluator): void;
  parser(parserName: ParserName, parser: CreateParser<ParserOptions, Parser<ParserReport>>): void;
  taskReporter(taskName: TaskName, report: TaskReporter): void;
  task(task: Task): void;
}
```

The rest of the PR is largely updating tests so they're compliant, and removing other bits of oclif that remained.


packages/core/src/types/cli.ts
```diff
@@ -25,3 +28,25 @@ export type RunOptions = {
...
+}
+
+export interface RegistrationProvider {
+  actions(taskName: TaskName, evaluate: TaskActionsEvaluator): void;
+  parser(parserName: ParserName, parser: CreateParser<ParserOptions, Parser<ParserReport>>): void;
+  taskReporter(taskName: TaskName, report: TaskReporter): void;
+  task(task: Task): void;
+}
+
+export interface RegistrationProviderOptions {
```
rwjblue, line 44: Is this duplicated in packages/cli/src/api/registration-provider.ts?

rwjblue (approved):

packages/core/src/types/cli.ts
scalvert: It's not meant to be. The other should be removed (I moved the types here to use them in multiple places). Will fix.

scalvert: Fixed.

PR #934, opened 2021-04-14
Rename reporters to formatters and associated types/code

scalvert:
Renames reporters to formatters to help unify the language around 'formatting output'. 

Also removes a few other remnants that were no longer used, specifically generator templates that are deprecated.


carakessler (approved):

carakessler:
We talked offline about fast follows for this regarding the api and the `format` option, will be coming in subsequent PRs

PR #935, opened 2021-04-14
Removes format from RunOptions to clean up Node API

scalvert:
As discussed, cleans up `RunOptions` to exclude `format`.


carakessler (approved):
thanks for the clean up!

PR #939, opened 2021-04-15
Refactor format options to align more closely with formatters

carakessler:
This change accomplishes the following:
1. --verbose option removed
2. stdout formatter renamed -> summary
3. verbose formatter renamed -> pretty
4. output-file param is respected by summary output


scalvert:
This is such a great change. Thanks for doing this, @carakessler!

PR #940, opened 2021-04-16
Pulling out cli-ux

carakessler:
We got cli-ux from oclif and we weren't using it for much. In preparation for a future change that writes the pretty format to a file, removing cli-ux in favor of just calling process.stdout.write directly, and using a lightweight cli table util.

Additionally, made the replacement for the ui object a class because previously there were arbitrary functions being called on the class that weren't causing failures in the compiler, and type checking wasnt being respected


packages/core/src/utils/stdout-formatter-utils.ts
```diff
@@ -43,67 +43,98 @@ export function calculateSectionBar(
   };
 }
 
-export const ui: typeof ux & { [key: string]: any } = Object.assign(ux, {
+export class StdoutFormatter {
```
scalvert, line 46: Since we just discussed yesterday about cleaning up the use of `*Formatter` classes, do you think we could rename this to something else so it's not confused with an actual formatter class? How about `StdoutWriter` or `ConsoleWriter`, since it's job is writing?

Also, can we just rename the filename to match the class name, whatever that may be?

scalvert (changes requested):

packages/core/src/utils/stdout-formatter-utils.ts
carakessler: 1000% good call

carakessler: fixing now

scalvert (approved):

PR #941, opened 2021-04-19
Adding functionality for pretty formatter to write to a file

carakessler:
With this refactor, the ConsoleWriter can operate in two modes. 1) write to console directly and 2) append content that wouldve been written to console to a string.
This way, the pretty formatter can write its contents to a txt file, instead of just printing to console.

This change makes it so that no two cli options are mutually exclusive! All format options work with the ouput-file option


packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
```diff
@@ -48,34 +47,15 @@ describe('eslint-disable-task', () => {
       })
     ).run();
 
-    expect(result.sort()).toMatchInlineSnapshot(`
```
carakessler, line 51: this test was hella flaky due to the random ordering of the JSON. future refactorings with the sarif matcher should fix this, but wanted to put in a patch for now and sort the results so they didnt come back in random orderings each time they were run

scalvert: Yep, this is why I'm working so hard right now on this `@microsoft/jest-sarif` matcher library!

packages/cli/__tests__/cli-test.ts
```diff
@@ -162,14 +162,27 @@ describe('cli-test', () => {
       join(project.baseDir, 'my-checkup-file.json'),
     ]);
 
-    let outputPath = result.stdout.trim();
+    let output = result.stdout.trim();
+    let outputPath = output.split(':')[1].trim(); // output will be a string followed by ':' then the file path
```
scalvert, line 166: I think you could just extend the regex below, or in fact wouldn't it work as is right now?

packages/cli/__tests__/cli-test.ts
```diff
@@ -162,14 +162,27 @@ describe('cli-test', () => {
...
     expect(existsSync(outputPath)).toEqual(true);
 
     unlinkSync(outputPath);
   });
 
+  it('should output a txt file in a custom directory if the pretty format and output-file options are provided', async () => {
+    let result = await run(['run', '.', '--format', 'pretty', `--output-file`, 'my-checkup-file']);
+
+    let output = result.stdout.trim();
+    let outputPath = output.split(':')[1].trim(); // output will be a string followed by ':' then the file path
```
scalvert, line 178: Same as above?

packages/cli/src/formatters/file-writer.ts
```diff
@@ -0,0 +1,60 @@
...
+  outputFile: string = DEFAULT_OUTPUT_FILENAME
+) {
+  let outputPath = getOutputPath(cwd, outputFile, 'txt');
+
+  writeFileSync(outputPath, stripAnsi(result));
+
+  consoleWriter.blankLine();
+  consoleWriter.log('Results have been saved to the following file:');
+  consoleWriter.log(yellow(outputPath));
+}
```
scalvert, line 38: Any reason not to just combine these into one, and pass in the extension?

packages/cli/src/formatters/formatter-utils.ts
```diff
@@ -1,9 +1,7 @@
 import { Action, CheckupMetadata, ConsoleWriter } from '@checkup/core';
 import { yellow, bold } from 'chalk';
 
-let consoleWriter = new ConsoleWriter();
-
-export function renderActions(actions: Action[]): void {
+export function renderActions(actions: Action[], consoleWriter: ConsoleWriter): void {
```
scalvert, line 4: Same comment re: `FormatOptions`.

packages/cli/src/formatters/formatter-utils.ts
```diff
@@ -13,7 +11,7 @@ export function renderActions(actions: Action[]): void {
   }
 }
 
-export function renderInfo(info: CheckupMetadata) {
+export function renderInfo(info: CheckupMetadata, consoleWriter: ConsoleWriter) {
```
scalvert, line 14: Same comment re: `FormatOptions`.

packages/cli/src/formatters/formatter-utils.ts
```diff
@@ -41,7 +39,7 @@ export function renderInfo(info: CheckupMetadata) {
   consoleWriter.blankLine();
 }
 
-export function renderLinesOfCode(info: CheckupMetadata) {
+export function renderLinesOfCode(info: CheckupMetadata, consoleWriter: ConsoleWriter) {
```
scalvert, line 42: Same comment re: `FormatOptions`.

packages/cli/src/formatters/formatter-utils.ts
```diff
@@ -55,7 +53,7 @@ export function renderLinesOfCode(info: CheckupMetadata) {
   consoleWriter.blankLine();
 }
 
-export function renderCLIInfo(info: CheckupMetadata) {
+export function renderCLIInfo(info: CheckupMetadata, consoleWriter: ConsoleWriter) {
```
scalvert, line 56: Same comment re: `FormatOptions`.

scalvert (changes requested):

packages/cli/src/formatters/file-writer.ts
carakessler: there is some duplication here, but the writing calls a different function (`writeFileSync` vs `writeJsonSync`, and we need to call `stripAnsi` for the txt one. I call pull the consoleWriting out into another function tho to remove that duplication

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
```diff
@@ -8,8 +8,7 @@ import {
 } from '@checkup/core';
 import { Result } from 'sarif';
 
-export function format(taskResults: Result[]) {
-  let consoleWriter = new ConsoleWriter();
+export function format(taskResults: Result[], consoleWriter: ConsoleWriter) {
```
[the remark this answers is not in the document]
carakessler: I understand the value of adding some abstraction here to prevent having to change all these signatures if we need to pass more data, especially since we're knocking on the door of 1.0.0, and these changes would be backwards incompat to other plugins' formatters. 

My only concern with what you proposed here is that the `ConsoleWriter` isn't really a `FormatOption`. The writer is in fact NOT optional ;) 
What do you think of calling the interface `FormatTools` or `FormatUtils` instead?

packages/cli/__tests__/cli-test.ts
carakessler: Imo its cleaner and simpler to just split on the `:` than include the whole output string in the regex here. Additionally, if the prose of the output string changes, we wont have to change the test - we only care what happens after that output string (which is the file path). Thoughts?

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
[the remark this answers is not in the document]
scalvert: It is options, depending on what perspective you have. For us, we provide the writer option, which would be required in the interface. Since we're passing the value into this function, it'd always be 'filled in'. 

If you're that opposed to options, can we call it args and `FormatArgs`?

packages/cli/__tests__/cli-test.ts
scalvert: But you don't need to include the string in the regex. The regex is `/^(.*)\/my-checkup-file.json/`, which would match any char preceding the path. 

`Whatever we want this to be can go here: some-path/my-checkup-file.json` matches right now...

packages/cli/src/formatters/file-writer.ts
scalvert: Why not just have a single function they both use, and each calls that? You can have a param to `stripAnsi` that's a bool?

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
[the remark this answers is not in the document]
carakessler: I like that, making the change now

packages/cli/src/formatters/file-writer.ts
carakessler: If I call `writeFileSync` with JSON, I need to JSON.stringify it, which removes the nice formatting and such. I still collapsed the two functions and just added some conditionals inside one writer function called `writeResultFile`

packages/cli/__tests__/cli-test.ts
carakessler: Ah the regex was failing because there is now a newline and `.` doesnt count those. The regex works if i change it to `(/^([\S\s]*)\/my-checkup-file.json/);`, but then if I want to check the files existence I still think the best way would be to split at the `:`, unless you have a better idea for that?

scalvert: If you want to check for the file, you can change the regex to specifically match the group, right? The reason I'd rather do this is doing splits and trims is very messy and error prone, and we've been trying really hard to stabilize the tests. I think modifying the regex could allow you to absorb more of that variability than doing string manipulations.

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
```diff
@@ -3,18 +3,17 @@ import {
...
   renderEmptyResult,
 } from '@checkup/core';
 import { Result } from 'sarif';
 
-export function format(taskResults: Result[]) {
-  let consoleWriter = new ConsoleWriter();
-  consoleWriter.section(taskResults[0].properties?.taskDisplayName, () => {
-    consoleWriter.log(
-      `${consoleWriter.emphasize('Octane Violations')}: ${sumOccurrences(taskResults)}`
+export function format(taskResults: Result[], formatArgs: FormatArgs) {
```
scalvert, line 11: Somewhat of a nitpick, but I'd probably just name this `args`, so that the subsequent call is a bit cleaner. I don't feel strongly about this, but I'd personally change it.

packages/cli/__tests__/cli-test.ts
carakessler: Sorry not a regex wizard here but there is a lot of path before the `my-checkup-file.json` we need to capture, and its all dynamic. Do you know how to do that? I can do some googling I suppose

scalvert: Yes I do! Can you give me a pattern to match, like an example?

carakessler:
```
Results have been saved to the following file:
/private/var/folders/w6/8sjlbd7s3sjgj_s643kyhft0000bs0/T/tmp-40681ycH4rhsExFPi/checkup-app/my-checkup-file.json
```

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
carakessler: call the type `args` or just the parameter?

packages/cli/__tests__/cli-test.ts
scalvert: Let's just split on the newline character and test the file path from that.

packages/checkup-plugin-ember/src/formatters/ember-octane-migration-status-formatter.ts
scalvert: Just the param. That'd be an odd type name... :)

scalvert (approved):

PR #942, opened 2021-04-20
Bumping eslint-plugin-unicorn

carakessler:
There are a bunch of new rules here, I turned off a few I just don't agree with, open to potentially turning off a few more if there are concerns!

### DIFF
--- a/packages/cli/src/api/checkup-task-runner.ts
+++ b/packages/cli/src/api/checkup-task-runner.ts
@@ -146,6 +146,8 @@ export default class CheckupTaskRunner {
       let evaluators = getRegisteredActions();
 
       for (let [taskName, evaluator] of evaluators) {
+        // https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1193
+        // eslint-disable-next-line unicorn/no-array-callback-reference
         let task = this.tasks.find(taskName);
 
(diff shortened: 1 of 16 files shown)


carakessler:
Once I merge #941 i will rebase and the failures in tests will be fixed. I will then have to fix all the new problems #941 will introduce and then will merge this

packages/cli/src/api/checkup-task-runner.ts
```diff
@@ -146,6 +146,8 @@ export default class CheckupTaskRunner {
       let evaluators = getRegisteredActions();
 
       for (let [taskName, evaluator] of evaluators) {
+        // https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1193
+        // eslint-disable-next-line unicorn/no-array-callback-reference
         let task = this.tasks.find(taskName);
```
scalvert, line 151: We _could_ rename this to `findTask`, so as to not confuse static analyzers.

scalvert (approved):

packages/cli/src/api/checkup-task-runner.ts
carakessler: yeah that makes sense, ill do it in a sep PR

PR #956, opened 2021-04-26
Fixes paths parsing to include multiple paths

scalvert:
Fixes incorrectly parsing only a single path when multiple paths are passed.


carakessler:
Can we add a test for this case?

scalvert:
Yep, working on it, hence the WIP.

scalvert:
I think we do have a test for it, BTW, but just confirming.

carakessler:
If we had a test wouldn't it have been failing previously though?

scalvert:
Let's add a more targeted test when I work on fixing up the expansion code.

carakessler (approved):

PR #957, opened 2021-04-27
Improve type definition for error kind map

evansolomon:
By typing the map based on the enum used in its keys you get better feedback when indexing it. This also helps the compiler understand the resulting value may be undefined and forces that condition to be handled instead of failing with something useless like"undefined is not a function".

An alternative would be to define ERROR_BY_KIND keys for the remaining ErrorKind values, in which case the resulting value would never be undefined.


scalvert:
Thanks for this, @evansolomon. A nice improvement.

scalvert (approved):

PR #963, opened 2021-04-28
Adding support to pass in a config inline to checkup

carakessler:
Renamed `config` param to `config-path`; `config` param now accepts an inline config.

### DIFF
--- a/packages/cli/__tests__/cli-test.ts
+++ b/packages/cli/__tests__/cli-test.ts
@@ -62,7 +62,8 @@ describe('cli-test', () => {
             --help           Show help  [boolean]
             --version        Show version number  [boolean]
         -e, --exclude-paths  Paths to exclude from checkup. If paths are provided via command line and via checkup config, command line paths will be used.  [array]
-        -c, --config         Use this configuration, overriding .checkuprc if present.
+        -c, --config-path    Use the configuration found at this path, overriding .checkuprc if present.
+            --config         Use this configuration, overriding .checkuprc if present.
         -d, --cwd            The path referring to the root directory that Checkup will run in  [default: (default)]
             --category       Runs specific tasks specified by category. Can be used multiple times.  [array]
             --group          Runs specific tasks specified by group. Can be used multiple times.  [array]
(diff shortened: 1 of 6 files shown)


packages/cli/__tests__/cli-test.ts
```diff
@@ -60,7 +60,8 @@ describe('cli-test', () => {
             --help           Show help  [boolean]
             --version        Show version number  [boolean]
         -e, --exclude-paths  Paths to exclude from checkup. If paths are provided via command line and via checkup config, command line paths will be used.  [array]
-        -c, --config         Use this configuration, overriding .checkuprc if present.
+        -c, --config-path    Use the configuration found at this path, overriding .checkuprc if present.
```
scalvert, line 63: It may be useful to just default this to `.checkuprc`, and expect it's going to be used throughout.

packages/cli/src/checkup.ts
```diff
@@ -92,7 +97,8 @@ checkup <command> [options]`
         let taskRunner = new CheckupTaskRunner({
           paths,
           excludePaths: argv.excludePaths as string[],
-          config: argv.config as string,
+          config: argv.config as CheckupConfig,
+          configPath: argv.configPath as string,
```
scalvert, line 101: Just an aside - we should really consider typing our yargs arguments do we don't have to cast like this.

scalvert (approved):
Overall seems great. Thanks @carakessler!

packages/cli/__tests__/cli-test.ts
carakessler: adding!

packages/core/src/config.ts
```diff
@@ -26,13 +26,13 @@ export const DEFAULT_CONFIG: CheckupConfig = {
   tasks: {},
 };
 
-export function getConfigPath(path: string = '') {
-  return join(resolve(path), '.checkuprc');
+export function getConfigPath(cwd: string = '', configPath: string = '.checkuprc') {
```
scalvert, line 29: Is there a need to default this here? Would it not just get the default from `options.configPath`?

scalvert (changes requested):

packages/core/src/config.ts
carakessler: writeConfig needs it

scalvert: You mean it's called from two places, and the latter needs it since it requires it when writing (generating a new config). That seems fine.

scalvert (approved):

PR #964, opened 2021-04-28
Making paths an optional param in RunOptions, defaulting to `.`

carakessler:
Giving paths a default value in RunOptions, used by CheckupTaskRunner

### DIFF
--- a/packages/core/src/types/cli.ts
+++ b/packages/core/src/types/cli.ts
@@ -9,7 +9,7 @@ export type RunOptions = {
   excludePaths?: string[];
   groups?: string[];
   listTasks?: boolean;
-  paths: string[];
+  paths?: string[];
   tasks?: string[];
 };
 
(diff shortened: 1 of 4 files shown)


packages/core/src/types/cli.ts
```diff
@@ -9,7 +9,7 @@ export type RunOptions = {
   excludePaths?: string[];
   groups?: string[];
   listTasks?: boolean;
-  paths: string[];
+  paths?: string[];
```
scalvert, line 12: Sorry, why are we making `paths` optional? We should always require a path, right?

scalvert (changes requested):

packages/core/src/types/cli.ts
carakessler: i thought we discussed the path being optional and defaulting it to `.`?

scalvert: Just defaulting it, not the optional bit (unless I missed that). 

What does it mean to operate on no paths?

scalvert: Also I think we need paths to be an array, in this case of one element.

scalvert: We resolved this offline.

scalvert (approved):

PR #983, opened 2021-04-30
Adding @microsoft/jest-sarif for log/result validation

scalvert:
This replaces all the non-deterministic tests using snapshots.


carakessler (approved):
so excited about this!!!!

PR #986, opened 2021-05-04
Fixing bug in config code

carakessler:
There was a bug introduced in the previous change to support inline configs that broke checkup finding a config at root/.checkuprc by default. This fixes that, and cleans up some of the naming to be more straightforward

### DIFF
--- a/packages/core/__tests__/config-test.ts
+++ b/packages/core/__tests__/config-test.ts
@@ -2,11 +2,11 @@ import fetch from 'node-fetch';
 import { readJsonSync, writeJsonSync, writeFileSync } from 'fs-extra';
 import { createTmpDir } from '@checkup/test-helpers';
 import {
-  getConfigPath,
+  resolveConfigPath,
   readConfig,
   writeConfig,
   parseConfigTuple,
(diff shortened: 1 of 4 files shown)


packages/core/__tests__/config-test.ts
```diff
@@ -80,14 +80,21 @@ describe('config', () => {
...
       let config = readConfig(configPath!);
 
       expect(config).toEqual(REMOTE_CONFIG);
     });
 
+    it('returns default config path if it is not defined', async () => {
+      // eslint-disable-next-line unicorn/no-useless-undefined
+      let configPath = await getConfigPath(undefined);
+
+      expect(configPath).toBe(`${process.cwd()}/.checkuprc`);
```
scalvert, line 95: Can we not use `process.cwd()` for tests? I'd prefer we use tmp dirs. Can we swap to that?

scalvert (changes requested):

packages/core/__tests__/config-test.ts
carakessler: done

scalvert (approved):

PR #1021, opened 2021-06-03
Adds addResult API to base-task

scalvert:
This is one step closer to improving the result building inside Checkup's Task class. This change adds a new API to the `BaseTask` class, which enables Task authors to add results in a more seamless way. This method will ensure the following:

1. The correct SARIF `Result` properties are populated
2. The SARIF `Rule` metadata is populated for this `Result` (future additions will normalize the rules added to ensure maximum SARIF log compression)

API Signature:

```ts
addResult(
    messageText: string,
    kind: TaskResultKind,
    level: TaskResultLevel,
    options?: TaskResultOptions
  ): RequiredResult;
```

Usage example:

```ts
class FakeTask extends BaseTask {
  taskName = 'my-fake';
  taskDisplayName = 'Fake';
  description = 'description';
  category = 'foo';

  run() {
    // gather some data

    this.addResult('This is a fake message', 'fail', 'error');
  }
}
```

Will populate the SARIF log like so:

```json
[
  {
    "kind": "informational",
    "level": "note",
    "message": {
      "text": "The is a fake message",
    },
    "properties": {
      "category": "foo",
      "taskDisplayName": "Fake",
    },
    "ruleId": "my-fake",
    "ruleIndex": 0,
  },
]
```


packages/core/src/types/checkup-log.ts
```diff
@@ -0,0 +1,18 @@
+import { SetRequired } from 'type-fest';
+import { Run, Result } from 'sarif';
+import { RunOptions } from '../types/cli';
+import { TaskListError, Action } from '../types/tasks';
+import { CheckupConfig } from '../types/config';
+import { FilePathArray } from '../utils/file-path-array';
+
+export type RequiredRun = SetRequired<Run, 'tool' | 'results'>;
+export type RequiredResult = SetRequired<Result, 'message' | 'ruleId' | 'kind' | 'level'>;
```
carakessler, line 9: is the `RequiredFoo` a standard naming convention? semantically feels a little confusing to me, but i may be missing something. I also realize these existed in a previous PR but i may have missed it/not reviewed

packages/core/src/base-task.ts
```diff
@@ -67,4 +88,55 @@ export default abstract class BaseTask {
...
+   * @param level One of a fixed set of strings that specify the severity level of the result
+   * @param location Specifies a location where the result occurred
+   * @param properties A property bag named properties, which stores additional values on the result
+   */
+  addResult(
+    messageText: string,
+    kind: TaskResultKind,
+    level: TaskResultLevel,
+    location?: TaskResultLocation,
+    properties?: TaskResultProperties
```
carakessler, line 106: this is super clean!!

packages/core/src/data/sarif-log-builder.ts
```diff
@@ -44,19 +18,16 @@ export default class SarifLogBuilder {
       $schema: 'https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json',
       runs: this.runs,
     };
-
-    // A sarifLog object SHALL contain a property named runs (§3.13.4)
-    this.addRun();
```
carakessler, line 49: why arent we adding a run anymore upon initalizing the log builder? the extra step seems unnecessary

carakessler (approved):
looking real good!!!! just a few qq

packages/core/src/types/checkup-log.ts
scalvert: It's not standard, no. What's confusing about it? What would you name a type that had optionals that are now required, but that would reflect the original type?

packages/core/src/data/sarif-log-builder.ts
scalvert: Because Rob correctly mentioned that if we do this, we're locking people into the specific run implementation rather than allowing them to add the run themselves. This is akin to tightly coupling the implementation, and/or assuming that people always want the initial run to have the provided props. If they don't, it's just a huge PITA to have to override the props within it. Does that make sense?

rwjblue (approved):

packages/core/__tests__/__utils__/tmp-dir.ts
```diff
@@ -0,0 +1,6 @@
+import { realpathSync } from 'fs';
+import tmp = require('tmp');
```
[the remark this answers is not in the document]
scalvert: True. Just force of habit.

packages/core/src/base-task.ts
scalvert: 🧹

packages/core/src/types/checkup-log.ts
scalvert: I'm going to merge this through but am happy to revisit naming after if you have some good ideas!

PR #1039, opened 2021-08-03
Implement Ink formatter

scalvert:


packages/checkup-formatter-pretty/src/components/index.ts
```diff
@@ -0,0 +1,16 @@
...
+  return new Map([
+    ['table', Table],
+    ['list', List],
+    ['listItem', ListItem],
+  ]);
+}
+
+export { List } from './list';
+export { Table } from './table';
+export { ListItem } from './list-item';
```
scalvert, line 16: Not sure why we're re-exporting these here, @zhanwang626. I know it's because we want them to be available to consumers, but I'm not sure this is the right way to do that for this package. We could simply import from the specific file in the top level index.ts.

packages/checkup-formatter-pretty/src/types/formatter.ts
```diff
@@ -0,0 +1,14 @@
...
+  json = 'json',
+  pretty = 'pretty',
+}
+
+export interface FormatterOptions {
+  cwd: string;
+  format: OutputFormat;
+  outputFile?: string;
+  log: Log;
+}
```
scalvert, line 14: These should not be in here. They're types that are exposed in either core or cli, and those packages should supply these types.

packages/checkup-formatter-pretty/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "allowJs": true,
+    "rootDir": "src",
+    "esModuleInterop": true
+  },
+  "references": [
+    {
+      "path": "../core"
+    },
+    {
+      "path": "../test-helpers"
```
scalvert, line 15: Does this use `test-helpers`?

packages/checkup-formatter-pretty/src/components/index.ts
zhanwang626: Okay, will remove the export here.

packages/checkup-formatter-pretty/tsconfig.json
zhanwang626: No, will remove this.

PR #1041, opened 2021-08-04
Adding custom formatter package

scalvert:
Adds the new custom formatter package that includes the Ink implementation.


packages/checkup-formatter-pretty/__tests__/summary-test.tsx
```diff
@@ -0,0 +1,1980 @@
+import * as React from 'react';
+import { Log } from 'sarif';
+import { render } from 'ink-testing-library';
+import { Summary } from '../src/components/summary';
+
+const result: Log = {
```
scalvert, line 6: We should probably move this test fixture to a separate file in order for this test to be more readable.

packages/checkup-formatter-pretty/src/pretty.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Box } from 'ink';
+import { getComponents } from './components';
+
+const Pretty = ({ data }: { data: string[] }) => {
```
scalvert, line 5: We should probably figure out a different name for this. I want to capture the idea that this is the top-level component, but right now just calling it `Pretty` doesn't capture that. Can we rename this to something closer to what we mean? Like, `PrettyFormatter`?

packages/checkup-formatter-pretty/__tests__/summary-test.tsx
zhanwang626: Will remove all the summary component related stuff.

zhanwang626 (approved):

PR #1049, opened 2021-08-19
Update checkup-formatter-pretty to output string

zhanwang626:
Changes the pretty formatter to output a string vs. writing to process.stdout directly.

Tasks included:
- [x] Remove write output file function from checkup-pretty-formatter
- [x] Update checkup-formatter-pretty to output string
- [x] Add tests for checkup-formatter-pretty format function
- [x] Implement an interface for custom components in the checkup formatter pretty.


packages/checkup-formatter-pretty/__tests__/formatter-test.ts
```diff
@@ -0,0 +1,53 @@
...
+  pretty = 'pretty',
+}
+
+describe('Test Pretty formatter', () => {
+  it('can generate string from format', async () => {
+    const log = readJsonSync(resolve(__dirname, './__fixtures__/checkup-result.sarif'));
+    const logParser = new CheckupLogParser(log);
+    const options: FormatterOptions = {
+      cwd: '',
+      format: OutputFormat.pretty,
```
scalvert, line 19: We should no longer have an `OutputFormat.pretty` enum value, since we're externalizing that format.

zhanwang626: Yeah, make sense.

zhanwang626:
In [index.ts](https://github.com/checkupjs/checkup/blob/c3492c5a2f4efc80e5d066e9bcecaed6b47db505/packages/checkup-formatter-pretty/src/index.ts#L36), I have insert two log info to check the stdout value:
```ts
  format(logParser: CheckupLogParser) {
    const stdout = new Stdout();
    console.log('Before render:', stdout);
    render(React.createElement(pretty, { logParser }), { stdout: stdout as any });
    console.log('After render:', stdout);
    return stdout.lastFrame();
  }
```
In local test, before/after render are different:
```
Before render: Stdout {
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  frames: [],
  write: [Function (anonymous)],
  lastFrame: [Function (anonymous)],
  [Symbol(kCapture)]: false
}

After render: Stdout {
  _events: [Object: null prototype] { resize: [Function (anonymous)] },
  _eventsCount: 1,
  _maxListeners: undefined,
  frames: [
    'Checkup report generated for travis v0.0.1  (1797 files analyzed)\n' +
      'This project is 9 years old, with 1448 active days, 5983 commits and 1667 files\n' +
      '\n' +
      '\n' +
      'lines of code 101513\n' +
      '\x1B[38;2;24;211;234m■■■■■■■■■■■■■■■■■■■■■■■■■ js (49161)\x1B[39m\n' +
      '\x1B[38;2;143;25;238m■■■■■■■■■■■■ svg (24112)\x1B[39m\n' +
...
```
But in [CI](https://github.com/checkupjs/checkup/runs/3405094688?check_suite_focus=true), they are the same, it seems `render` function from ink did not work correctly.
```
Before render: Stdout {
PASS __tests__/pretty-test.tsx
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  frames: [],
  write: [Function],
  lastFrame: [Function],
  [Symbol(kCapture)]: false
}
After render: Stdout {
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  frames: [],
  write: [Function],
  lastFrame: [Function],
  [Symbol(kCapture)]: false
}
```

packages/checkup-formatter-pretty/src/types/index.ts
```diff
@@ -3,3 +3,9 @@ export interface BarData {
   value: number;
   total: number;
 }
+
+export interface Options {
```
scalvert, line 7: It's a bit odd to be redeclaring this type here. I'd prefer if we extract that type to `@checkup/core` if we need to use it outside of CLI (it can go in the cli.ts types file).

packages/checkup-formatter-pretty/src/index.ts
```diff
@@ -1,89 +1,43 @@
...
-  outputFile: string = DEFAULT_OUTPUT_FILENAME
-): string {
-  let fileType: 'sarif' | 'txt' = typeof result === 'string' ? 'txt' : 'sarif';
-  let outputPath = getOutputPath(cwd, outputFile, fileType);
-
-  if (fileType === 'txt') {
-    writeFileSync(outputPath, stripAnsi(result));
-  } else {
-    writeJsonSync(outputPath, result);
+class Stdout extends EventEmitter {
```
scalvert, line 9: Can we not just create a `NodeJS.WriteStream` instead of a class that extends from EventEmitter?

scalvert: I guess ink does it this way https://github.com/vadimdemedes/ink/blob/259edd47968081b23b176b7a7cfecc5e5757845d/test/helpers/create-stdout.ts#L5 (at least for testing)

packages/checkup-formatter-pretty/src/types/index.ts
zhanwang626: the FormatterOptions in `@checkup/core` are using type `OutputFormat` which is enum
```ts
export interface FormatterOptions {
  cwd: string;
  format: OutputFormat;
  outputFile?: string;
}
export enum OutputFormat {
  summary = 'summary',
  json = 'json',
}
```

scalvert (approved):

PR #1060, opened 2021-08-31
Change other formatters to return strings

scalvert:
This is the second part of #1049, and changes the other two formatters to return strings.

Of note, the summary formatter still uses the `outputFile` option, mainly as its output isn't meant to be stored in a file. If we feel this is confusing or inconsistent, we can change it.

Note: This PR also includes some fixes to the Octane migration status task that were found during testing.

### DIFF
--- a/packages/cli/src/checkup.ts
+++ b/packages/cli/src/checkup.ts
@@ -1,11 +1,12 @@
 import * as yargs from 'yargs';
 import * as ora from 'ora';
+import { yellow } from 'chalk';
 import { OutputFormat, ConsoleWriter, CheckupConfig } from '@checkup/core';
-
 import CheckupTaskRunner from './api/checkup-task-runner';
 import Generator from './api/generator';
 import { getFormatter } from './formatters/get-formatter';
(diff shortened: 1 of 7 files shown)


packages/cli/src/checkup.ts
```diff
@@ -136,7 +137,23 @@ checkup <command> [options]`
           });
 
           spinner.stop();
-          formatter.format(log);
+          let output = formatter.format(log);
+
+          if (output) {
+            if (argv.outputFile) {
+              let resultFilePath = writeResultFile(
```
zhanwang626, line 144: Since we writeResultFile here, should we remove `writeResultsToFile` function from `summary` and `json`?

scalvert: The `summary` formatter still uses it, but it would be good to drop it down from the Base formatter. I'm not even sure we need the `BaseFormatter` anymore, but I can think on that.

scalvert: It looks like that function isn't in the `BaseFormatter` anyway, so I think there's no change needed here.

Issue #1053, opened 2021-08-30
Ensure all tasks are retrofitted with their component type for rendering


scalvert:
The `checkup-formatter-pretty` formatter allows you to render output using components, and further allows you to specify the component in the task's metadata. We need to ensure each task has a component property added to the rule's metadata, and that property is used to output the respective task.

eg. The following is a task that migrates my stuff

```ts
"driver": {                       
  "name": "CodeScaner",
  "rules": [
    {                             
      "id": "my-migration",
      "shortDescription": {
        "text": "A migration for my things"
      },
      "properties": {
        "component": "migration"
      }
    }
  ]
}
```

PR #1061, opened 2021-08-31
Retrofit components

zhanwang626:
Ensure all tasks are retrofitted with their component type for rendering
issue: #1053

### DIFF
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -26,12 +26,14 @@ export default class EmberDependenciesTask extends BaseTask implements Task {
               startLine: dependency.startLine,
               startColumn: dependency.startColumn,
             },
-            properties: {
-              packageName: dependency.packageName,
-              packageVersion: dependency.packageVersion,
-              latestVersion: dependency.latestVersion,
-              type: dependency.type,
-              component: 'table',
+            rule: {
+              properties: {
+                packageName: dependency.packageName,
+                packageVersion: dependency.packageVersion,
+                latestVersion: dependency.latestVersion,
+                type: dependency.type,
+                component: 'table',
+              },
             },
           }
         );
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -31,6 +31,7 @@ export default class EmberDependenciesTask extends BaseTask implements Task {
               packageVersion: dependency.packageVersion,
               latestVersion: dependency.latestVersion,
               type: dependency.type,
+              component: 'table',
             },
           }
         );
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -26,13 +26,15 @@ export default class EmberDependenciesTask extends BaseTask implements Task {
               startLine: dependency.startLine,
               startColumn: dependency.startColumn,
             },
+            properties: {
+              packageName: dependency.packageName,
+              packageVersion: dependency.packageVersion,
+              latestVersion: dependency.latestVersion,
+              type: dependency.type,
+            },
             rule: {
               properties: {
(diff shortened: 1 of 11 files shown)


scalvert:
This is a great start! One thing we need to do is ensure that all `component` definitions are under a parent `rule` property. That way when we call `addResult`, the `rule` is merged with the other rule metadata.

Example:

```ts
      this.addResult('some message',
        'review',
        'warning',
        {
          location: {
            uri: filePath,
            startColumn: column,
            startLine: line,
          },
          properties: {
            //...
          },
		  rule: {
		    properties: {
              component: 'table',
            }
          }
        }
      );
```

zhanwang626:
> This is a great start! One thing we need to do is ensure that all `component` definitions are under a parent `rule` property. That way when we call `addResult`, the `rule` is merged with the other rule metadata.
> 
> Example:
> 
> ```ts
>       this.addResult('some message',
>         'review',
>         'warning',
>         {
>           location: {
>             uri: filePath,
>             startColumn: column,
>             startLine: line,
>           },
>           properties: {
>             // properties that are added to each result
>           },
> 		  rule: {
> 		    properties: { // properties that are added to the rule metadata
>               component: 'table',
>             }
>           }
>         }
>       );
> ```

Yeah, I just make sure all of the component definitions are under a parent rule property and updated this PR.

packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
```diff
@@ -26,11 +26,14 @@ export default class EmberDependenciesTask extends BaseTask implements Task {
...
-              packageName: dependency.packageName,
-              packageVersion: dependency.packageVersion,
-              latestVersion: dependency.latestVersion,
-              type: dependency.type,
+            rule: {
+              properties: {
+                packageName: dependency.packageName,
+                packageVersion: dependency.packageVersion,
+                latestVersion: dependency.latestVersion,
+                type: dependency.type,
```
scalvert, line 34: These properties shouldn't be under `rule.properties`. Only the `component` definition, but the others should be back up one level where they were before. They pertain to the result, not the rule.

packages/checkup-plugin-ember/src/tasks/ember-octane-migration-status-task.ts
```diff
@@ -226,6 +226,7 @@ export default class EmberOctaneMigrationStatusTask extends BaseTask implements
             properties: {
               parentRuleID: this.taskName,
               taskDisplayName: `Ember Octane Migration | ${ruleMetadata.feature}`,
+              component: 'table',
```
scalvert, line 229: We should wait until #1060 is merged as this definition has changed.

packages/checkup-plugin-javascript/src/tasks/outdated-dependencies-task.ts
```diff
@@ -27,11 +27,14 @@ export default class OutdatedDependenciesTask extends BaseTask implements Task {
...
-              packageName: dependency.packageName,
-              packageVersion: dependency.packageVersion,
-              latestVersion: dependency.latestVersion,
-              type: dependency.type,
+            rule: {
+              properties: {
+                packageName: dependency.packageName,
+                packageVersion: dependency.packageVersion,
+                latestVersion: dependency.latestVersion,
+                type: dependency.type,
```
scalvert, line 35: Similar to the other comment above, these don't belong here. They're part of the top-level `properties`, not the `rule.properties`.

scalvert (changes requested):
As an aside, I'm not sure we should be setting everything to use `table`. I wouldn't expect that'd be the primary output we'd want for all tasks. We should discuss this, but we should likely chose something a bit richer.

packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
zhanwang626: Got it.

zhanwang626:
> As an aside, I'm not sure we should be setting everything to use `table`. I wouldn't expect that'd be the primary output we'd want for all tasks. We should discuss this, but we should likely chose something a bit richer.

Sure, we can set the default component to `task-result-list` which can display more details about nested infomation of a task result.

scalvert (approved):

Issue #1056, opened 2021-08-30
Update list and bar component to render task result correctly.


scalvert:
Also rename current bar to sectioned-bar.

PR #1064, opened 2021-09-01
Update list and bar component to render task result correctly.

zhanwang626:
#1056 

- [x] Rename `task-result-bar`, `task-result-list` to `bar` and `list`.

- [x] Rename current bar component to `sectioned-bar`.

- [x] Let `bar` and `list` component to render task result correctly.

- [x] Update all task's metadata to have `list` component in `rule.properties.component`.


packages/checkup-formatter-pretty/src/components/list.tsx
```diff
@@ -0,0 +1,23 @@
...
+export const List: React.FC<{ taskResult: RuleResults }> = ({ taskResult }) => {
+  return (
+    <Box flexDirection="column">
+      <Text underline>{formatRuleId(taskResult.rule.id)}</Text>
+      <Text>Total: {taskResult.results.length}</Text>
+    </Box>
+  );
+};
+
+function formatRuleId(ruleId: string): string {
```
scalvert, line 18: We should use `taskResult.rule.properties.taskDisplayName` instead of this. If your log doesn't have it, you should regenerate the test fixture because this has been added.

scalvert (approved):

PR #1077, opened 2021-09-10
Converts built-in LOC info to task

scalvert:
Moves the lines of code data to a task vs metadata. This allows consumers to invoke it separately, or not invoke it at all.


packages/checkup-formatter-pretty/src/components/table.tsx
```diff
@@ -8,6 +8,10 @@ import { TaskDisplayName } from '../sub-components/task-display-name';
 import { getOptions } from '../get-options';
 
 type TableOptions = {
+  sumBy: {
+    findGroupBy: string;
+    sumValueBy: string;
+  };
```
scalvert, line 14: This capability was added to the table component so it was possible to summarize data by configured table column property.

packages/checkup-formatter-pretty/src/components/table.tsx
```diff
@@ -26,17 +30,37 @@ export const Table: React.FC<{ taskResult: RuleResults }> = ({ taskResult }) =>
 
 function buildTableData(taskResult: RuleResults): any[] {
   let { rule, results } = taskResult;
-  let { rows } = getOptions<TableOptions>(rule);
+  let { rows, sumBy } = getOptions<TableOptions>(rule);
 
-  let dataRows = results.map((result: Result) => {
-    let rowData: Record<string, any> = {};
+  return sumBy !== undefined
```
scalvert, line 35: I feel like this could _definitely_ use some documentation and comments, as it's not super easy to grok what is exactly going on in here. The gist is we're doing some data formatting that's entirely declaratively driven via the Task's rule metadata. This allows us to generically format the data for output.

PR #1094, opened 2021-09-20
Fixes plugin generator not generating in custom subdirectory

scalvert:


packages/cli/src/api/generator.ts
```diff
@@ -44,23 +44,15 @@ export default class Generator {
...
     try {
-      await new Promise<void>((resolve, reject) => {
-        env.run(`checkup:${type}`, generatorOptions, (err: Error | null) => {
-          if (err) {
-            reject(err);
-          } else {
-            // this is ugly, but I couldn't find the correct configuration to ignore
-            // generating the yeoman repository directory in the cwd
-            let yoRepoPath = join(this.options.path, '.yo-repository');
+      // @ts-ignore
```
scalvert, line 47: This is required due to a mismatch of @types modules that are transitive dependencies. The version of `yeoman-environment` _does_ have a `run` implementation that returns a promise, and the callback variant is actively deprecated. However, our `@types/yeoman-environment` version clashes with this, indicating that the `run` method doesn't return a promise.

PR #1104, opened 2021-09-23
Add export to @checkup/cli for writeResultFile

carakessler:
This exports the writeResultFile from cli, and adds writeResultsToFile as an optional function on the Formatter interface, to expose it for formatters that populate it

### DIFF
--- a/packages/core/src/types/cli.ts
+++ b/packages/core/src/types/cli.ts
@@ -1,3 +1,4 @@
+import { Log } from 'sarif';
 import CheckupLogParser from '../data/checkup-log-parser';
 import {
   Task,
@@ -45,6 +46,7 @@ export interface RegisterableTaskList {
 
 export interface Formatter {
   format(logParser: CheckupLogParser): string;
+  writeResultsToFile?(log: Log): void;
 }
 
 export interface FormatterCtor {
--- a/packages/core/src/types/cli.ts
+++ b/packages/core/src/types/cli.ts
@@ -1,4 +1,3 @@
-import { Log } from 'sarif';
 import CheckupLogParser from '../data/checkup-log-parser';
 import {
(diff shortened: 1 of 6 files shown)


packages/core/src/types/cli.ts
```diff
@@ -45,6 +46,7 @@ export interface RegisterableTaskList {
 
 export interface Formatter {
   format(logParser: CheckupLogParser): string;
+  writeResultsToFile?(log: Log): void;
```
scalvert, line 49: As we discussed, I don't want to do this as we specifically removed this functionality to simplify the formatters, and to make their interface match that of eslint (run a formatter and return a string). This makes the formatters very easy to test, and removes them writing to stdout.

Could we just keep this PR to the export of the `writeResultFile` function?

scalvert (changes requested):

packages/core/src/types/cli.ts
carakessler: what's the point of having the functions in the formatters if they can't be called? i assumed our conversation was scoped to adding the writing stuff as an option and requiring for matters to handle it. i don't see what is wrong with optionally allowing formatters to expose this functionality (especially since it's exist and can't be invoked as written)

scalvert: As I mentioned before, we don't want formatters to necessarily write their own outputFile. That's what the option is for. Summary is a unique exception to this, but for others implementing formatters we'd prefer they generate whatever output then need, and let the CLI handle the rest. For the Node API case, they can just import the util and write the file themselves.

scalvert: If it's ok with you, I'll just resolve this for now. I'm happy to have this discussion in a follow up!!

PR #1105, opened 2021-09-23
deps: Updating fixturify-project

### LINKED ISSUE #1103: Bump to latest node-fixturify-project to fix esModuleInterop issue
Found an issue this evening which is a combination of a bunch of factors:

1. `compilerOptions.esModuleInterop` is a "viral" option: if a package upstream sets it and uses its features (specifically `export = <some name>` or `import <some name> = require(...)`), all downstreams are then potentially exposed to it as well.
2. An older version of `node-fixturify-project` (at least through 2.x, but fixed as of 4.x, unsure about 3.x) both set `compilerOptions.esModuleInterop: true` and also used the `export =` functionality.
3. This package currently uses `node-fixturify-project@2` in its test helpers and imports it using the form which *requires* interop support: `import ProjectFixture from 'node-fixturify-project';`.

This is, to be clear, a *super* easy mess to end up in. Indeed: the defaults generated by `tsc --init` and indeed recommended by the TS team make it almost impossible *not* to fall into this trap! There's even a TS LS "fixer" which suggests changing from the compatible-with-non-interop-downstream `import ProjectFixture = require('node-fixturify-project');` to the incompatible-with-non-interop ES module syntax! 😭 

Happily, the solution here is straightforward (and *hopefully* also easy?): to update to the latest version of `'node-fixturify-project'`, which has switched to a named export. 🎉

(All of this is why the still-open RFC for SemVer for types in TS requires conforming projects to set `esModuleInterop: false`.)

I’m also going to open an issue and PR on the upstream to flip *off* `esModuleInterop` so that this won’t even accidentally affect downstream going forward, and will hopefully have a similar PR for this one, though it depends on how many of the upstreams for this require it!

---

(I wrote all of this up rather than *just* doing the fix so we have a quick reference to point to in the future for stuff like this. I may also turn it into a blog post at some point b/c I see it come up pretty often.)

scalvert:
Fixes #1103


scalvert:
cc/ @chriskrycho

packages/checkup-plugin-javascript/package.json
```diff
@@ -37,7 +37,7 @@
     "rimraf": "^3.0.2"
   },
   "engines": {
-    "node": ">= 12.11.*"
+    "node": ">= 12.22.*"
```
scalvert, line 40: This was necessary due to an engines mismatch.

chriskrycho:
💙 thanks!

PR #1107, opened 2021-09-23
Ensuring global execution defaults to local if present

scalvert:
If checkup is invoked as a global, we want to ensure that it's first using a locally installed version, and falls back to using a global if present.

eg.

**Invoking checkup when checkup is installed locally:**

```sh-session
checkup run .
```

Should use the **local** version of `@checkup/cli`. 

**Invoking checkup globally when checkup is installed locally:**

```sh-session
checkup run .
```

Should use the **local** version of `@checkup/cli`. 

**Invoking checkup globally when checkup is _not_ installed locally:**

Either via `npx`:

```sh-session
npx @checkup/cli run .
```

Or via global install:

```sh-session
checkup run .
```

Should use the **global** version of `@checkup/cli`.

### DIFF
--- a/packages/cli/bin/checkup.js
+++ b/packages/cli/bin/checkup.js
@@ -2,8 +2,12 @@
 
 require('v8-compile-cache');
 
-const { run } = require('../lib/checkup');
+const importLocal = require('import-local');
+const checkup = importLocal(require.resolve('../lib/checkup.js'));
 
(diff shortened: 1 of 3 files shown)


rwjblue (approved):

scalvert:
Unsure why my lockfile isn't changed...

PR #1108, opened 2021-09-25
Extracting rule metadata from addResult

scalvert:
The `addResult` API is a cleaner addition to the `Task` class, and helps for adding valid SARIF to the resulting log. One awkward part of that API was it being overloaded to _also_ add the rule metadata. The awkwardness centered around a few things:

- it felt like a side-effect to invoke `addRule` inside `addResult`
- if no results were found for a task, the rule metadata would never be populated
- calling `addResult` multiple times, which is pretty common, would result in `addRule` being also called multiple times unnecessarily. While there was logic in `addRule` to ensure a new rule wasn't added every time it was called if one already existed, this still felt like unnecessary execution
- it made it hard for users to understand the relationship between the rule metadata and a result. There really wasn't any relationship other than the rule metadata represented static information added to the log about the `Task`, not the result.

Based on the above, it felt more correct to extract the `addRule` method out of `addResult` and require each Task to individually invoke it to set things up correctly in the SARIF log. While it seems somewhat redundant to require all Tasks to call this boilerplate, it gives Task authors an opportunity to add additional metadata to the rule, if they desire.

### DIFF
--- a/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
+++ b/packages/checkup-plugin-ember/src/tasks/ember-dependencies-task.ts
@@ -1,5 +1,5 @@
 import { join } from 'path';
-import { BaseTask, Task, DependencyAnalyzer } from '@checkup/core';
+import { BaseTask, Task, TaskContext, DependencyAnalyzer } from '@checkup/core';
 import { Result } from 'sarif';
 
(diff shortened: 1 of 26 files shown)


scalvert:
@rwjblue this is what we'd discussed the other day, as a final step to 1.0.

PR #1112, opened 2021-09-29
Fixes Pretty formatter outputing empty results for all tasks

scalvert:
Due to changes to the way we build rule metadata, there are now differences in the way we want to output results. 

Formerly the pretty formatter would try to indicate when no results were present, which helped verify to the user that the task ran with no results. Since the change to _always_ allow rule metadata, and the addition of tracking executed tasks separately, we needed to additionally filter when outputting the 'no results' info to only do so for executed tasks.


carakessler (approved):
cute!

PR #1113, opened 2021-09-30
Fixes bug in get-paths where subdir segments were truncated

scalvert:
There was a bug in the get-paths module where it was incorrectly truncating the paths for directories (`/somedir`). Specifically, that directory would explicitly be removed from the full path, resulting in an invalid path on disc.


scalvert:
Going to merge this through as it's a gnarly one.

PR #1135, opened 2021-11-10
Adds valid-esm-package-task

scalvert:
Adds a new validation task for validating ESM packages.

Validation output:

```sh-session
Checkup ran the following task(s) successfully:

Valid ESM Package
=================
Validation passed
  ✔ Contains "type" property set to "module" in package.json
  ✔ Does not contain a "main" property in package.json
  ✔ Contains an "exports" property set to a valid value in package.json
  ✔ Requires at least Node 13.2.0 in the "engines" property of package.json
  ✔ Should not have "use strict" in any JavaScript files
```

Uses heuristic from https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c to determine if a package is a valid ESM package.

### DIFF
--- a/packages/core/src/base-validation-task.ts
+++ b/packages/core/src/base-validation-task.ts
@@ -7,7 +7,7 @@ export type ValidationResult = {
 };
 
 export default abstract class BaseValidationTask extends BaseTask {
-  validationSteps: Map<string, () => ValidationResult>;
+  validationSteps: Map<string, () => ValidationResult | Promise<ValidationResult>>;
 
   /**
    * Creates a new instance of a validation Task.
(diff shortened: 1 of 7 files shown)


packages/core/src/base-validation-task.ts
```diff
@@ -49,11 +52,11 @@ export default abstract class BaseValidationTask extends BaseTask {
    *
    * @returns A map of messages and ValidationResult objects
    */
-  validate() {
+  async validate() {
```
scalvert, line 55: This is technically a breaking change, but I wince at the thought of another major already. Open to any alternatives.

PR #1137, opened 2021-11-15
[enhancement] Adding non-fatal errors to tasks

carakessler:
## Rationale: 
Previously, when a single file could not be parsed correctly, the entire eslint-disable task would fail. We want to notify our consumers when a file can not be parsed, or when any error occurs when analyzing a single file, but we dont necessarily want that to cause the whole task to fail. It is now up to the author of a task whether they just want to throw an error when a failure is caught, or just add a nonFatalError, which are all collected at the end and pushed to the `toolExecutionNotifications`.

## Implementation: 

Added new public function `addNonFatalError` to base-task, and collected the nonFatalErrors in task-list with the other (fatal) errors. This approach pushes the responsibility of handling failures in the analyzers onto the task authors, giving them full control of how they want to handle what happens when a file cant be analyzed (for some tasks, I imagine failure to read/parse a file to be fatal, and for the desired behavior to be a task erroring out completely). With this change, made required changes to eslint-disable task (the only task using the AstAnalyzer right now).

### DIFF
--- a/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
+++ b/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
@@ -108,9 +108,8 @@ describe('eslint-disable-task', () => {
     await task.run();
 
     expect(task.nonFatalErrors).toHaveLength(1);
-    expect(task.nonFatalErrors![0]).toMatchInlineSnapshot(
-      // eslint-disable-next-line jest/no-interpolation-in-snapshots
-      `[SyntaxError: Error occurred at ${project.root}/foo/error-file.js. Unexpected token (9:18)]`
+    expect(task.nonFatalErrors[0].message).toEqual(
+      `Error occurred at ${project.root}/foo/error-file.js. Unexpected token (9:18)`
     );
   });
 });
--- a/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
+++ b/packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
@@ -79,4 +79,38 @@ describe('eslint-disable-task', () => {
       }
     `);
   });
+
+  it('captures a non-fatal error for nonparsable files', async () => {
+    project.files['error-file.js'] = `
+      let config;
+
+      try {
+        const metaName = '{{MODULE_PREFIX}}/config/environment';
+        const rawConfig = document.querySelector('meta[name="' + metaName + '"]').getAttribute('content');
+        config = JSON.parse(unescape(rawConfig));
+      } catch(err) {
+        config = {{CONFIG_CONTENTS}};
+      }
+
+      export default config;
+    `;
+
+    project.writeSync();
+
+    const task = new EslintDisableTask(
+      pluginName,
+      getTaskContext({
+        paths: project.filePaths,
+        options: { cwd: project.baseDir },
+      })
+    );
+
+    await task.run();
+
+    expect(task.nonFatalErrors).toHaveLength(1);
+    expect(task.nonFatalErrors![0]).toMatchInlineSnapshot(
+      // eslint-disable-next-line jest/no-interpolation-in-snapshots
+      `[SyntaxError: Error occurred at ${project.root}/foo/error-file.js. Unexpected token (9:18)]`
+    );
+  });
 });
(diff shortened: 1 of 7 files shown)


packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
```diff
@@ -79,4 +79,38 @@ describe('eslint-disable-task', () => {
...
+      getTaskContext({
+        paths: project.filePaths,
+        options: { cwd: project.baseDir },
+      })
+    );
+
+    await task.run();
+
+    expect(task.nonFatalErrors).toHaveLength(1);
+    expect(task.nonFatalErrors![0]).toMatchInlineSnapshot(
```
scalvert, line 111: I think you can just use `toEqual` in this case, which would allow you to avoid the eslint disable (ironic this is in a PR related to eslint-disable stuff).

packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
```diff
@@ -63,61 +63,66 @@ export default class EslintDisableTask extends BaseTask implements Task {
...
+              visit,
+              {
+                parser: require('recast/parsers/babel'),
+              }
+            );
+            analyzer.analyze(accumulator.visitors);
+            data.push(...accumulator.data);
+          } catch (error) {
+            (error as Error).message = `Error occurred at ${filePath}. ${(error as Error).message}`;
+            this.addNonFatalError(error as Error);
```
scalvert, line 120: I'd probably just throw a new error instead, rather than mutating the current error's message.

packages/cli/src/task-list.ts
```diff
@@ -185,6 +185,7 @@ export default class TaskListImpl implements RegisterableTaskList {
 
     try {
       result = await this._runTask(task);
+      this.addErrors(task.fullyQualifiedTaskName, task.nonFatalErrors);
```
scalvert, line 188: Should this always be called? What if there are no fatal errors?

scalvert (changes requested):

packages/checkup-plugin-javascript/__tests__/eslint-disable-task-test.ts
carakessler: lol meta!! good call, fixing

packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
carakessler: if i do that, i would lose the stack trace, no?

packages/cli/src/task-list.ts
carakessler: then it would be called with an empty array, and nothing would happen. I can wrap this in an `if (task.nonFatalErrors.length > 0)` if you'd prefer

packages/cli/src/task-list.ts
```diff
@@ -282,4 +284,15 @@ export default class TaskListImpl implements RegisterableTaskList {
...
+  private addErrors(taskName: TaskName, errors: Error[]) {
+    this._errors.push(
+      ...errors.map((error) => {
+        return {
+          error,
+          taskName,
+        };
+      })
+    );
+  }
```
scalvert, line 297: Let's change this to 

```suggestion
  private addErrors(taskName: TaskName, errors: Error | Error[]) {
    if (!Array.isArray(errors)) {
      errors = [errors];
    }
  
    this._errors.push(
      ...errors.map((error) => {
        return {
          error,
          taskName,
        };
      })
    );
  }
```

and then just delete the `addError` function.

carakessler: lol just did exactly that but no mutation and with a ternary so #oneline ;) updating

packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
scalvert: This is true, and since we're wanting to preserve that stack to ultimately output it in the SARIF log, we should keep it this way.

packages/cli/src/task-list.ts
scalvert: No it's fine, so long as it's not actually adding any errors, which it doesn't. Thanks for clarifying.

scalvert (approved):

PR #1176, opened 2022-03-01
chore: bump ember-template-lint to v4

lbdm44:


packages/core/src/analyzers/ember-template-lint-analyzer.ts
```diff
@@ -11,8 +12,6 @@ import {
 import { TaskConfig } from '../types/config';
 import { mergeLintConfig } from '../utils/merge-lint-config';
 
-const TemplateLinter = require('ember-template-lint');
```
lbdm44, line 14: `ember-template-lint` does not produce/provide any type definitions for itself, because of this TS won't let us use import statements as is

scalvert: I think we just need to add a declaration file in `./types/ember-template-lint.d.ts` that reads:

```ts
declare module 'ember-template-lint';
```

lbdm44: was looking into this, its still complaining even if i add one, but i have a ton of other stuff I'm trying to work though and may have been in the wrong tmux session, will circle back around to this later today

lbdm44: Alright, got the typing to work, but jest is balking now with:
```
/Users/lemiller/projects/checkup/node_modules/ember-template-lint/lib/index.js:2
    export { default } from './linter.js';
    ^^^^^^

    SyntaxError: Unexpected token 'export'
```
which ive seen before, need to look at some personal projects to see how i fixed those

scalvert: ya this is because we need to run jest in `esm` mode via `jest`/`ts-jest`.

scalvert: I think we're going to have an issue here with trying to use esm in cjs. It's not really compatible per se.

lbdm44: By default, yes, the cononical solution is to use `babel-jest` to transpile during test runs, this project is using `ts-jest` for typescript files but doesnt have a fallback for `js` at the moment.

I did some digging, we would need to do some of the following:

1. modify the `transformIgnorePatterns` jest option to _not_ ignore `ember-template-lint`
2. follow the basic setup for babel-jest via jest's docs
3. upgrade to use node@lts since ember-template-lint uses the new `node:` import protocol that is only supported in 16+
4. fiddle with the babel setup to get imports not messing up

At the current moment, i do not have time to push this along further, this was meant to be a WIP for someone at work to look at and crib off of.

scalvert: Ya. Makes sense. Ultimately I'm wanting to jettison jest in favor of vitest, which overall manages the usage of ESM within tests much better (there's no need for intermediaries like ts-jest). Perhaps that can help with this work.

Also, I do have a branch that I'm working on that contains a bunch of these changes already, but is somewhat blocked by TS/ESM mismatch until a newer version of TS is released that better supports TS/ESM interop.

Thanks for starting this though!

PR #1210, opened 2022-05-10
Converting from commonjs to esm

scalvert:


scalvert:
Going to pause the `npm` migration as there's a type resolution conflict that occurs within `npm`'s `overrides` that doesn't occur in `yarn`'s `resolutions`. Until we can normalize the version numbers across packages, we should keep on `yarn`.

PR #1236, opened 2022-06-21
Adding task error output to indicate tasks ran, but with errors

scalvert:
Adds output to indicate whether the task ran with non-fatal errors or not. 

This was a regression when we swapped over from standard output writing to Ink. This change includes some updates/changes to that error output.

Example output:

Screen Shot 2022-06-21 at 9 57 53 AM

### DIFF
--- a/packages/ui/__tests__/components/bar-test.tsx
+++ b/packages/ui/__tests__/components/bar-test.tsx
@@ -6,7 +6,7 @@ import { CheckupLogParser, dirname } from '@checkup/core';
 import stripAnsi from 'strip-ansi';
 import { Bar } from '../../src/components/Bar';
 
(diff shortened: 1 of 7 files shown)


zhanwang626 (approved):
Looks good to me! I think it might be better that we add a unit test for TaskError component?

scalvert:
@zhanwang626 I agree. There's a number of components that need tests, and I'd like to add those all in a separate PR. I can add the `TaskErrors` test in this PR now though!

scalvert:
@zhanwang626 I added a quick sanity check test for `TaskErrors`.

zhanwang626 (approved):
Great work, thanks!

PR #1243, opened 2022-06-23
Refactors path expansion for CLI inputs

scalvert:
Updates file path expansion logic to simplify and correctly observe path exclusions. Fixes a long standing bug with Checkup incorrectly including `dot` paths/directories.

### DIFF
--- a/packages/cli/__tests__/cli-test.ts
+++ b/packages/cli/__tests__/cli-test.ts
@@ -533,8 +533,8 @@ describe('cli-test', () => {
     result = await run(['run', '.']);
     let unfiltered = result.stdout;
 
-    expect(filtered).toContain('4 files analyzed');
-    expect(unfiltered).toContain('6 files analyzed');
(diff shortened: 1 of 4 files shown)


packages/cli/__tests__/cli-test.ts
```diff
@@ -533,8 +533,8 @@ describe('cli-test', () => {
     result = await run(['run', '.']);
     let unfiltered = result.stdout;
 
-    expect(filtered).toContain('4 files analyzed');
-    expect(unfiltered).toContain('6 files analyzed');
+    expect(filtered).toContain('2 files analyzed');
+    expect(unfiltered).toContain('4 files analyzed');
```
scalvert, line 537: This applies to all of the tests that were changed, but the new path expansion logic fixes an issue whereby Checkup was inadvertently including files that should have been excluded. This optimization has had the knock on effect of changing the test output.

PR #1248, opened 2022-06-29
Recursively ignore build files

jalexakos:
Certain repositories use things such as yarn workspaces, which nest addons in packages. Those build folders would not be ignored before. This fixes that.


scalvert (approved):
Thanks for adding this!

PR #1352, opened 2023-06-27
Update eslint-disable-task to disregard *.d.ts files and use recast typescript parser

chrisrng:
- Avoid running checkup on autogenerated files
- Fix parsing issue when given a typescript file by using the typescript parser

Example error:
```
{"message":{"text":"javascript/eslint-disables generated an error during execution"},"level":"error","associatedRule":{"id":"javascript/eslint-disables"},"exception":{"message":"Error occurred at file.ts. Unexpected token, expected \",\" (24:29)"
```

### DIFF
--- a/packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
+++ b/packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
@@ -42,7 +42,7 @@ export default class EslintDisableTask extends BaseTask implements Task {
   }
 
   async run(): Promise<Result[]> {
-    let esLintablePaths = this.context.paths.filterByGlob(['**/*.js', '**/!(*.d).ts']);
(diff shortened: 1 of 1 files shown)


packages/checkup-plugin-javascript/src/tasks/eslint-disable-task.ts
```diff
@@ -42,7 +42,7 @@ export default class EslintDisableTask extends BaseTask implements Task {
   }
 
   async run(): Promise<Result[]> {
-    let esLintablePaths = this.context.paths.filterByGlob(['**/*.js', '**/*.ts']);
+    let esLintablePaths = this.context.paths.filterByGlob(['**/*.js']);
```
scalvert, line 45: Should we just swap to the TS parser rather than remove this?

scalvert (changes requested):

scalvert (approved):

